Issue #1, opened 2015-09-02
Problems with non-json data.


ballPointPenguin:
I was attempting to use ember-json for a POST with FormData (Content-Type:multipart/form-data) for a file upload operation. Unfortunately, ember-ajax overrides the contentType and also passes the data through JSON.stringify.

https://github.com/ember-cli/ember-ajax/blob/master/addon/services/ajax.js#L132

Perhaps there can be an option to disable this feature or maybe disable it automatically if options.dataType or options.contentType are set to something other than json.

If not, then the documentation should clearly state that this service is designed ONLY for json data.

taras:
I agree that `ember-ajax` should not enforce payload to be in JSON format. I rarely work with non JSON data, so I don't have much personal experience with that uses case. I would like to take some time on deciding how to add this to the library(PRs are welcome, as I see you've done already in #2 ).

abuiles:
This is fixed by #5

ballPointPenguin:
thanks, @abuiles

PR #2, opened 2015-09-02
Check if data is FormData and don't treat like JSON if it is.

ballPointPenguin:
Note: I don't expect to get merged this as-is. I'm putting it here as a 'strawman' to even see if anybody thinks non-json content types should be supported and how that support should be included. Some options might include:

1. Only check if data is FormData, otherwise treat as JSON (current state of my fork -- Band-Aid for my immediate problem).
2. Allow user to opt-out of json with an additional param to ajax() call, otherwise treat as JSON.
3. Check if data is valid JSON and handle it as such only if it is.
4. Check for existing options such as 'dataType' and/or 'contentType' and make assumptions from there.

### DIFF
--- a/addon/services/ajax.js
+++ b/addon/services/ajax.js
@@ -128,7 +128,14 @@ export default Ember.Service.extend({
     hash.dataType = 'json';
     hash.context = this;
 
-    if (hash.data && type !== 'GET') {
+    var formData;
+    if (hash.data && hash.data.constructor) {
+      if(hash.data.constructor.name === 'FormData') {
+        formData = true;
+      }
+    }
+
+    if (hash.data && type !== 'GET' && !formData) {
       hash.contentType = 'application/json; charset=utf-8';
       hash.data = JSON.stringify(hash.data);
     }


ballPointPenguin:
in response to #1

taras:
Thank you for the first PR :)

> Only check if data is FormData, otherwise treat as JSON (current state of my fork -- Band-Aid for my immediate problem).

It looks ok, I would do something like this

```javascript
import Ajax from 'ember-ajax/services/ajax';

export default Ajax.extend({
   /**
    @method ajaxOptions
    @private
    @param {String} url
    @param {String} type The request type GET, POST, PUT, DELETE etc.
    @param {Object} options
    @return {Object}
  */
  options(url, type, options) {
    var hash = options || {};
    hash.url = url;
    hash.type = type;
    hash.dataType = 'json';
    hash.context = this;

    if (hash.data && type !== 'GET' && !hasFormData(hash)) {
      hash.contentType = 'application/json; charset=utf-8';
      hash.data = JSON.stringify(hash.data);
    }

    var headers = get(this, 'headers');
    if (headers !== undefined) {
      hash.beforeSend = function (xhr) {
        Object.keys(headers).forEach((key) =>  xhr.setRequestHeader(key, headers[key]));
      };
    }

    return hash;
  },
});

function hasFormData(hash) {
    var formData;
    if (hash.data && hash.data.constructor) {
      if(hash.data.constructor.name === 'FormData') {
        return true;
      }
    }
  return false;
}
```

> Allow user to opt-out of json with an additional param to ajax() call, otherwise treat as JSON.

I'm thinking about splitting up the options method into 2 methods: `defaultHash` and `options`.

```javascript
 /**
    @method options
    @private
    @param {String} url
    @param {String} type The request type GET, POST, PUT, DELETE etc.
    @param {Object} options
    @return {Object}
  */
  options(url, type, options) {
    const hash = this.defaultHash(...arguments);

    if (hash.data && type !== 'GET') {
      hash.contentType = 'application/json; charset=utf-8';
      hash.data = JSON.stringify(hash.data);
    }

    return hash;
  },
   /**
    @method defaultHash
    @private
    @param {String} url
    @param {String} type The request type GET, POST, PUT, DELETE etc.
    @param {Object} options
    @return {Object}
  */
  defaultHash(url, type, options) {
    const hash = options || {};
    hash.url = url;
    hash.type = type;
    hash.dataType = 'json';
    hash.context = this;

    var headers = get(this, 'headers');
    if (headers !== undefined) {
      hash.beforeSend = function (xhr) {
        Object.keys(headers).forEach((key) =>  xhr.setRequestHeader(key, headers[key]));
      };
    }
   return hash;
  }
```

With this setup, you could do something like this:

```javascript
import Ajax from 'ember-ajax/services/ajax';

export default Ajax.extend({
   /**
    @method ajaxOptions
    @private
    @param {String} url
    @param {String} type The request type GET, POST, PUT, DELETE etc.
    @param {Object} options
    @return {Object}
  */
  options(url, type, options) {
    
    if (type === 'POST' && hasFormData(options)) {
      const hash = this.defaultHash(...arguments);
      // add your stuff here
      return hash;      
    } 
    
    return this._super(...arguments);
  },
});

function hasFormData(hash) {
    var formData;
    if (hash.data && hash.data.constructor) {
      if(hash.data.constructor.name === 'FormData') {
        return true;
      }
    }
  return false;
}
```

> Check if data is valid JSON and handle it as such only if it is.

I'd love to see what that could look like.

> Check for existing options such as 'dataType' and/or 'contentType' and make assumptions from there.

Let me know what your thoughts are about this as well.

ballPointPenguin:
I realize now that this is the way that ember-data does it (assumes all requests to be JSON). I was using jquery.ajax for these FormData (file upload) POST's, but that was problematic, especially for testing. I had expected ember-ajax to basically just wrap xhr request in Ember-y runloop and promise syntax and was very surprised to find it manipulating, in fact invalidating, my data. Seeing the parallel between the code in ember-ajax and in ember-data sheds more light on how / why this came to be :bulb:

taras:
Yes, this code was pulled from Ember Data. The goal is to create a single service for making AJAX requests. So, some of the original code needs to be generalized so to make it less Ember Data specific.

ballPointPenguin:
Anyway. I like the first and second options the best. Especially the second option (decouple the defaultHash function form the options function, giving the user great flexibility).

Checking if something is "valid JSON" is a bit dubious and may seems a bit draconian in cases where somebody is using less-than-valid JSON on purpose, or the validation logic is less-than-perfect.

Similarly, parsing and making assumptions from contentType and/or dataType could be problematic. contentTypes are numerous and subject to whimsy, and dataType is really concerned with what the server is going to return and not what is being sent.

taras:
One of the other things to consider is that we ultimately want to be able to provide [/stefanpenner/ember-fetch](/stefanpenner/ember-fetch) as an option. So whatever APIs are added here, ideally, we'd decouple it from jQuery.ajax implementation and just make it a convenient way making AJAX calls.

abuiles:
Superseded by #5

ballPointPenguin:
ok nice. FWIW, I'm not using my fork anymore, I'm actually just using the example you provided above for extending the service.

PR #5, opened 2015-09-25
[ENHANCEMENT] Don't make assumptions about dataType and payload.

abuiles:
Since the final goal of this is to become a service for making **AJAX** request, it would be a good idea not to make many assumptions about such request, **AJAX** does not necessarily mean  **json**. 

I'd like to propose we just expose the basic interface for **jQuery.ajax** without trying to be too smart about it. With this we also address issues like #2.

Also what about changing the the args of `service.request` to be ONLY `url` and `options`, I think this provides better ergonomics that the current implementation, for example if I want to do a get with query params I'd have to do something like `service.request(url, 'GET', {data: {foo: true}}` this could be reduce to `service.request(url, {data: {foo: true}}` since `GET` tends to be the default, and then if I need a different request type, then I could just specify that in the options (like `$.ajax`)

### DIFF
--- a/addon/services/ajax.js
+++ b/addon/services/ajax.js
@@ -125,13 +125,8 @@ export default Ember.Service.extend({
     var hash = options || {};
     hash.url = url;
     hash.type = type;
-    hash.dataType = 'json';
+    hash.dataType = hash.dataType || 'json';
     hash.context = this;
-
-    if (hash.data && type !== 'GET') {
-      hash.contentType = 'application/json; charset=utf-8';
-      hash.data = JSON.stringify(hash.data);
-    }
 
     var headers = get(this, 'headers');
     if (headers !== undefined) {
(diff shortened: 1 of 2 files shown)


taras:
Great simple solution. Thank you

abuiles:
@taras I'll open a second PR for the argument change, notice they are breaking changes.

taras:
> Also what about changing the the args of service.request to be ONLY url and options, I think this provides better ergonomics that the current implementation, for example if I want to do a get with query params I'd have to do something like service.request(url, 'GET', {data: {foo: true}} this could be reduce to service.request(url, {data: {foo: true}} since GET tends to be the default, and then if I need a different request type, then I could just specify that in the options (like $.ajax)

I agree. I think is a good idea. Can we make it backwards compatible?

devinus:
This would have been great if it then checked for `dataType` here https://github.com/ember-cli/ember-ajax/commit/63973c31115d192b5ca77b196f7a8c8c88282c10#diff-6ad1f65c87cc1996c63b65a765d84933L131 rather than removing it.

PR #6, opened 2015-09-29
[ENHANCEMENT] Explicitly default options type to GET

scottkidder:
Set the type on the options when it is not provided. This fixes the promise label seen in the inspector.


taras:
Looks good, thank you :)

abuiles:
let's just get rid of the second argument :) and be `ur`l and `options`

PR #15, opened 2015-10-22
Allow host to be set on ajax service

dasnixon:
Add documentation for host property with examples


taras:
This looks great. Thank you

PR #16, opened 2015-10-23
Add HTTP-verb methods on ajax service

dasnixon:
* `post(url, options)` - sets options.type to POST
* `del(url, options)` - sets options.type to DELETE
* `put(url, options)` - sets options.type to PUT

Added documentation and a few comments in code

### DIFF
--- a/addon/services/ajax.js
+++ b/addon/services/ajax.js
@@ -133,6 +133,29 @@ export default Ember.Service.extend({
     }, `ember-ajax: ${hash.type} to ${url}`);
   },
 
+  // calls `request()` but forces `options.type` to `POST`
+  post(url, options) {
+    return this.request(url, this._addTypeToOptionsFor(options, 'POST'));
+  },
+
+  // calls `request()` but forces `options.type` to `PUT`
+  put(url, options) {
+    return this.request(url, this._addTypeToOptionsFor(options, 'PUT'));
+  },
+
+  // calls `request()` but forces `options.type` to `DELETE`
+  del(url, options) {
+    return this.request(url, this._addTypeToOptionsFor(options, 'DELETE'));
+  },
+
+  // forcibly manipulates the options hash to include the HTTP method on the type key
+  _addTypeToOptionsFor(options, method) {
+    options = options || {};
+    delete options.type;
+    options.type = method;
+    return options;
+  },
+
   /**
     @method options
     @private
--- a/addon/services/ajax.js
+++ b/addon/services/ajax.js
@@ -151,7 +151,6 @@ export default Ember.Service.extend({
   // forcibly manipulates the options hash to include the HTTP method on the type key
   _addTypeToOptionsFor(options, method) {
     options = options || {};
-    delete options.type;
     options.type = method;
     return options;
   },
--- a/tests/unit/services/ajax-test.js
+++ b/tests/unit/services/ajax-test.js
@@ -152,9 +152,8 @@ test("post() promise label is correct", function(assert) {
   const postPromise = service.post(url, options);
   assert.equal(postPromise._label, 'ember-ajax: POST to /posts');
 
-  postPromise.then(function (response) {
-    assert.equal(response.post.title, title);
-    assert.equal(response.post.description, description);
+  return postPromise.then(function (response) {
+    assert.deepEqual(response.post, options.data.post);
   });
 });
 
@@ -177,10 +176,8 @@ test("put() promise label is correct", function(assert) {
   const putPromise = service.put(url, options);
   assert.equal(putPromise._label, 'ember-ajax: PUT to /posts/1');
 
-  putPromise.then(function (response) {
-    assert.equal(response.post.title, title);
-    assert.equal(response.post.description, description);
-    assert.equal(response.post.id, id);
+  return putPromise.then(function (response) {
+    assert.deepEqual(response.post, options.data.post);
   });
 });
 
@@ -195,8 +192,8 @@ test("del() promise label is correct", function(assert) {
   const delPromise = service.del(url);
   assert.equal(delPromise._label, 'ember-ajax: DELETE to /posts/1');
 
-  delPromise.then(function (response) {
-    assert.equal(response, {});
+  return delPromise.then(function (response) {
+    assert.deepEqual(response, {});
   });
 });
 
--- a/tests/unit/services/ajax-test.js
+++ b/tests/unit/services/ajax-test.js
@@ -115,8 +115,8 @@ test("options() type defaults to GET", function(assert) {
 
 test("request() promise label is correct", function(assert) {
   service = Service.create();
-  const url = '/posts';
-  const data = {
+  let url = '/posts';
+  let data = {
     type: 'POST',
     data: {
       post: { title: 'Title', description: 'Some description.' }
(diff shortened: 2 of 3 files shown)


MattNguyen:
+1

tests/unit/services/ajax-test.js
```diff
@@ -134,6 +134,49 @@ test("request() promise label is correct", function(assert) {
   assert.equal(postPromise._label, 'ember-ajax: POST to /posts');
 });
 
+test("post() promise label is correct", function(assert) {
+  service = Service.create();
+  const url = '/posts';
```
stefanpenner: these all seem like `let`.

`const` should be reserved for things that absolutely must stay constant, not `let` that  happen to remain constant.
Individuals should feel compelled to NOT change `const` -> `let`, over utilizing it degrades the value of this messaging.

tests/unit/services/ajax-test.js
```diff
@@ -134,6 +134,49 @@ test("request() promise label is correct", function(assert) {
...
+  const options = {
+    data: {
+      post: { title: 'Title', description: 'Some description.' }
+    }
+  };
+  const serverResponse = [200, { "Content-Type": "application/json" }, JSON.stringify(options.data)];
+
+  server.post(url, () => serverResponse);
+
+  const postPromise = service.post(url, options);
```
stefanpenner: we should actually validate the responses.

```js
return service.post(url, options).then( more assetions)
```

tests/unit/services/ajax-test.js
```diff
@@ -134,6 +134,49 @@ test("request() promise label is correct", function(assert) {
...
+  const options = {
+    data: {
+      post: { title: 'Title', description: 'Some description.' }
+    }
+  };
+  const serverResponse = [200, { "Content-Type": "application/json" }, JSON.stringify(options.data)];
+
+  server.put(url, () => serverResponse);
+
+  const putPromise = service.put(url, options);
```
stefanpenner: ^

tests/unit/services/ajax-test.js
```diff
@@ -134,6 +134,49 @@ test("request() promise label is correct", function(assert) {
...
+});
+
+test("del() promise label is correct", function(assert) {
+  service = Service.create();
+  const url = '/posts/1';
+  const serverResponse = [200, { "Content-Type": "application/json" }, JSON.stringify({})];
+
+  server.delete(url, () => serverResponse);
+
+  const delPromise = service.del(url);
```
stefanpenner: ^

tests/unit/services/ajax-test.js
elwayman02: As I understand it, `const` means "constant reference", not "constant value", @stefanpenner...is this wrong?

elwayman02: More specifically, I'm based that off the MDN description of `const`: 
> The const declaration creates a read-only reference to a value. It does not mean the value it holds is immutable, solely that the variable identifier can not be reassigned.

It seems like you are saying that we should treat `const` as if it is immutable, which is the opposite of what its documentation claims.

addon/services/ajax.js
```diff
@@ -133,6 +133,29 @@ export default Ember.Service.extend({
...
+
+  // calls `request()` but forces `options.type` to `DELETE`
+  del(url, options) {
+    return this.request(url, this._addTypeToOptionsFor(options, 'DELETE'));
+  },
+
+  // forcibly manipulates the options hash to include the HTTP method on the type key
+  _addTypeToOptionsFor(options, method) {
+    options = options || {};
+    delete options.type;
```
elwayman02: Why do you need to delete type right before you reassign it? Couldn't you just have line 155 on its own?

stefanpenner: ya no point.

tests/unit/services/ajax-test.js
stefanpenner: No, I specifically made no reference to immutability. Local vars that "happen" to not change, are not `const`, they are let. `const` is meant to convey, please do to rebind this given variable.

elwayman02: Can you give an example of when you would want to use `const`, then?

stefanpenner:
```

const USER_URL = 'http:...../'

export default Ember.Object.extend({

  actions: {
     save(user) {
        post(USER_URL + '/' + user.get('id'));
     }
  }
});
```

elwayman02: That's an example of something that, in any other language, would be considered immutable. Despite you not making a specific reference to immutability, you are still asserting that `const` should only be used in this type of scenario. Thus, I am confused why the ES6 spec bothered to introduce a keyword that is apparently *supposed* to convey this concept while specifically calling for a different implementation altogether. That seems relatively useless, does it not?

stefanpenner: let/const/var describing the variable binding, not the content.

const is immutable binding, let is block scoped binding, var is yolo binding.

stefanpenner: > That's an example of something that, in any other language, would be considered immutable.

I don't believe thats actually the case.

elwayman02: Ok. I don't know that I'm entirely convinced of the usefulness of that distinction, but it's not an unreasonable explanation. Thanks for clarifying, @stefanpenner!

stefanpenner: It's subtle, but has value.

dasnixon:
@elwayman02 @stefanpenner updated, `let`, me know your thoughts (get the pun). I think I used `let`/`const` more correctly?

tests/unit/services/ajax-test.js
```diff
@@ -127,13 +127,79 @@ test("request() promise label is correct", function(assert) {
...
+  };
+
+  const serverResponse = [200, { "Content-Type": "application/json" }, JSON.stringify(options.data)];
+
+  server.post(url, () => serverResponse);
+
+  const postPromise = service.post(url, options);
+  assert.equal(postPromise._label, 'ember-ajax: POST to /posts');
+
+  postPromise.then(function (response) {
```
stefanpenner: these need to return, to be async tests.

tests/unit/services/ajax-test.js
```diff
@@ -127,13 +127,79 @@ test("request() promise label is correct", function(assert) {
...
+  };
+
+  const serverResponse = [200, { "Content-Type": "application/json" }, JSON.stringify(options.data)];
+
+  server.put(url, () => serverResponse);
+
+  const putPromise = service.put(url, options);
+  assert.equal(putPromise._label, 'ember-ajax: PUT to /posts/1');
+
+  putPromise.then(function (response) {
```
stefanpenner: ^

tests/unit/services/ajax-test.js
```diff
@@ -127,13 +127,79 @@ test("request() promise label is correct", function(assert) {
...
+  let url = '/posts/1';
+
+  const serverResponse = [200, { "Content-Type": "application/json" }, JSON.stringify({})];
+
+  server.delete(url, () => serverResponse);
+
+  const delPromise = service.del(url);
+  assert.equal(delPromise._label, 'ember-ajax: DELETE to /posts/1');
+
+  delPromise.then(function (response) {
```
stefanpenner: ^

dasnixon:
@stefanpenner updated and used `deepEqual` to compare the `response` from the promise to the `options` object.

stefanpenner:
@dasnixon thank you kindly, LVGTM

@embersherpa r?

taras:
@stefanpenner I don't have an opinion on this, so if you think it's good then I'll merge and release it.

taras:
@dasnixon thank you for your contribution and @stefanpenner thank you for the code review :)

taras:
Just published 0.7.0 with this PR. Thank you :)

dasnixon:
Thanks I appreciate it!

PR #18, opened 2015-10-26
Use najax when DOM is not present

rondale-sc:
When the DOM isn't specified we can assume that this service is being
used within the context of Fastboot.  Unfortunately, Ember.$.ajax won't
do the correct thing.  Fastboot has created a service:

https://github.com/tildeio/ember-cli-fastboot/blob/master/app/initializers/server/ajax.js

That implements adapter#ajax in the context of node using `najax`

This checks to see if the DOM is present and if so behaves as it did
before, however when DOM isn't present it will assume Fastboot has
loaded najax and will call it passsing in the same `hash` parameters

### DIFF
--- a/addon/services/ajax.js
+++ b/addon/services/ajax.js
@@ -73,6 +73,7 @@ const {
   ```
 
 **/
+
 export default Ember.Service.extend({
 
   request(url, options) {
@@ -129,7 +130,11 @@ export default Ember.Service.extend({
          reject(error);
       };
 
-      Ember.$.ajax(hash);
+      if (typeof najax === 'undefined') {
+        Ember.$.ajax(hash);
+      } else {
+        najax(hash); // jshint ignore:line
+      }
     }, `ember-ajax: ${hash.type} to ${url}`);
   },
 


addon/services/ajax.js
```diff
@@ -73,6 +73,8 @@ const {
   ```
 
 **/
+const hasDOM = typeof document !== 'undefined';
```
stefanpenner: i read this as canHazDom :trollface:

stefanpenner:
LGTM, is there a reasonable way for us to get this under test? 

cc @tomdale i think you have WIP making addons test in node by default?

addon/services/ajax.js
stefanpenner: this actually isn't correct, as there is no document in webworkers

rondale-sc: @stefanpennar I can check for the existence of `najax`  (which I believe was rwjblue's original suggestion)

topaxi:
Anything to be done here? Can I help out?

taras:
@tomdale do you have any suggestions how how we can test this in node?

tomdale:
@topaxi @taras I'm hoping to have a first spike of a node-compatible service this week.

taras:
@tomdale awesome! look forward to it

tomdale:
I think this PR can be closed in favor of having addon users use something like [https://github.com/tomdale/ember-network](https://github.com/tomdale/ember-network).

alexlafroscia:
@tomdale: Cool. Ultimately, `ember-ajax` is going to be moving that way anyway, so I was thinking the same thing.

PR #32, opened 2015-12-16
Adds `namespace` option. Fixes #30

### LINKED ISSUE #30: Server API version
As of now, I am using the `host` option to configure the server API version.

```js
// app/services/ajax.js

import Ember from 'ember';
import AjaxService from 'ember-ajax/services/ajax';

export default AjaxService.extend({
  host: '/api/v1'
});

```

Is that ok?

manoharank:


manoharank:
cc @taras

taras:
@alexlafroscia can you please look at this. My only reservation was that it looks like we modified existing tests instead of adding new once. Is that the case?

alexlafroscia:
Funny, I was just about to comment and say that I looked it over and it looks pretty good to me, too.  I did have the same thought though -- why was the URL for those tests changed?  Also, as simple as they are, separate tests for the `_normalizePath` method might not be a bad idea.

alexlafroscia:
@manoharank can you rebase off of `master` and update your PR?

manoharank:
Git commits are collapsed. I will create a new PR.

manoharank:
Here is new PR #41

PR #34, opened 2016-01-05
Update usage documentation

### LINKED ISSUE #33: Docs: readme could mention jquery ajax api
It'd be helpful to mention the base library at the top of the readme, as some users will come to this having never used ic-ajax, and possibly without realising (or forgetting) it's jQuery based. 

[ic-ajax](https://github.com/instructure/ic-ajax) makes it clear near the start of the readme that
> This lib simply wraps jQuery.ajax with two exceptions: … Other than that, use request exactly like $.ajax.

I just tripped over the 'data' parameter like this:

```js
this.get('ajax').post('/admin/login', {email: email, password: password})
```
fills in the http request's password instead of putting it into the form data.

What I needed was 

```js
this.get('ajax').post('/admin/login', { data: {email: email, password: password} })
```

It's in the readme examples, but doesn't say what it is, and I thought it was structured data rather than request configuration.

alexlafroscia:
Adds information about configation taking the same format as `jQuery.ajax`.

Closes #33

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,21 +20,45 @@ export default Ember.Route.extend({
   model() {
     return this.get('ajax').request('/posts');
   }
-})
+});
 ```
 
 ## Ajax Service
 
+### Basic Usage
+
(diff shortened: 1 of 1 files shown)


taras:
LGTM, Thanks :)

PR #38, opened 2016-01-05
Add error types for 400 and 5XX errors

### LINKED ISSUE #36: Add missing Ajax failures
For example, 400 should throw BadRequest error.

alexlafroscia:
Adds two new errors types, `BadRequestError` and `ServerError`, which correspond to `400` and `5XX` errors, respectively.

Closes #36

### DIFF
--- a/addon/services/ajax.js
+++ b/addon/services/ajax.js
@@ -4,7 +4,9 @@ import {
   AjaxError,
   UnauthorizedError,
   InvalidError,
-  ForbiddenError
+  ForbiddenError,
+  BadRequestError,
+  ServerError
 } from '../errors';
 import parseResponseHeaders from '../utils/parse-response-headers';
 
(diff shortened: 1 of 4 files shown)


addon/services/ajax.js
```diff
@@ -225,6 +227,7 @@ export default Ember.Service.extend({
    @return {Object | DS.AdapterError} response
  */
  handleResponse(status, headers, payload) {
+   payload = payload || {};
```
alexlafroscia: I'm not 100% sure why, but when I was testing the `5XX` errors against a few different values, only the `510

taras:
LGTM; will wait for tests to pass and merge

addon/services/ajax.js
taras: > only the `510
Still writing?

alexlafroscia:
@taras I went ahead and merged this since the tests are passing.

taras:
great!

Issue #19, opened 2015-11-05
Documentation improvement


gitjeff05:
I searched for an answer to this, but I was wondering if you could clarify some things about this statement in the README.md:

> ic-ajax also wraps requests in Ember.run which is no longer necessary on Ember 1.13+.

Do you mean that you no longer have to wrap requests or callbacks? I see [in ic-ajax that they wrap the success callback in an Ember.run](https://github.com/instructure/ic-ajax/blob/master/lib/main.js#L105). Also, it looks like [ember-ajax library does the same thing](https://github.com/ember-cli/ember-ajax/blob/master/addon/make-promise.js#L17) if I am reading it right.

So why is this no longer necessary? Ember 1.13 docs still state that wrapping async callbacks with `Ember.run` is a good idea. I couldn't find anything in the Ember docs that stated this.

taras:
@gitjeff05 the code that you're pointing to was there to allow an upgrade path from `ic-ajax`. It should and will be removed before 1.0. It should throw a deprecation warnings when used. If it doesn't - it's a bug.

Here is an explanation from @krisselden 

<img width="669" alt="screen shot 2015-11-05 at 5 00 24 pm" src="https://cloud.githubusercontent.com/assets/74687/10983125/e260d0de-83de-11e5-9c08-bb62d0cf2f28.png">

gitjeff05:
> so Ember.run isn't something people every think about anymore.

I cannot wait for that day. However, now I am even more confused about the run loop, event loop and promises.

alexlafroscia:
@taras @gitjeff05 is this still an issue that needs a documentation update to fix?  Or is everything settled at this point?

taras:
We need to restore the [`Ember.run` wrapper](https://github.com/emberjs/data/blob/master/addon/adapters/rest.js#L857) because it allows us to use the [wait()](https://github.com/switchfly/ember-test-helpers/blob/master/lib/ember-test-helpers/wait.js) helper and update documentation to mention that we're wrapping in Runloop.

alexlafroscia:
Gotcha.  I'll try to take care of that then :)

alexlafroscia:
@taras since we did this:

> We need to restore the Ember.run wrapper because it allows us to use the wait() helper and update documentation to mention that we're wrapping in Runloop.

in #40, is this issue okay to close?

PR #40, opened 2016-01-06
Wrap request handlers in Ember run loop

alexlafroscia:
- Re-implements `Ember.run` wrapping of handler code, as mentioned here -> https://github.com/ember-cli/ember-ajax/issues/19#issuecomment-169190521
- Adds documentation regarding run loop wrapping, as well as other general cleanup of the Readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,10 +49,11 @@ export default Ember.Controller.extend({
 });
 ```
 
+In this example, `this.get('ajax').request()` will return a promise with the result of the request.  Your handler code inside `.then` or `.catch` will automatically be wrapped in an Ember run loop for maximum compatibility with Ember, right out of the box.
+
 ### HTTP-verbed methods
 
-You can skip setting the `method` or `type` keys in your `options` object when calling `request(url, options)` by
-instead calling `post(url, options)`, `put(url, options)`, `patch(url, options)` or `del(url, options)`.
+You can skip setting the `method` or `type` keys in your `options` object when calling `request(url, options)` by instead calling `post(url, options)`, `put(url, options)`, `patch(url, options)` or `del(url, options)`.
 
 ```js
 post('/posts', { data: { title: 'Ember' } })    // Makes a POST request to /posts
@@ -63,12 +64,9 @@ del('/posts/1')                                 // Makes a DELETE request to /po
 
 ### Custom Request Headers
 
-`ember-ajax` allows you to specify headers to be used with a request. This is
-especially helpful when you have a session service that provides an auth token
-that you have to include with the requests to authorize your requests.
+`ember-ajax` allows you to specify headers to be used with a request. This is especially helpful when you have a session service that provides an auth token that you have to include with the requests to authorize your requests.
 
-To include custom headers to be used with your requests, you can specify `headers`
-hash on the `Ajax Service`.
+To include custom headers to be used with your requests, you can specify `headers` hash on the `Ajax Service`.
 
 ```js
 // app/services/ajax.js
@@ -93,12 +91,9 @@ export default AjaxService.extend({
 
 ### Custom Host
 
-`ember-ajax` allows you to specify a host to be used with a request. This is
-especially helpful so you don't have to continually pass in the host along
-with the path, makes `request()` a bit cleaner.
+`ember-ajax` allows you to specify a host to be used with a request. This is especially helpful so you don't have to continually pass in the host along with the path, makes `request()` a bit cleaner.
 
-To include a custom host to be used with your requests, you can specify `host`
-property on the `Ajax Service`.
+To include a custom host to be used with your requests, you can specify `host` property on the `Ajax Service`.
 
 ```js
 // app/services/ajax.js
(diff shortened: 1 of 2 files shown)


addon/services/ajax.js
```diff
@@ -113,9 +113,9 @@ export default Ember.Service.extend({
         );
 
         if (response instanceof AjaxError) {
-          reject(response);
+          Ember.run.join(null, reject, response);
```
toranb: learning opportunity (for me) -why ember.run.join here instead of ember.run?

rwjblue: `Ember.run` will create a new run loop (even if one is already started), `Ember.run.join` will join an existing run loop or create a new one if one isn't underway.

toranb: @rwjblue thanks for the TL;DR

My deeper question is - how would this work in the following scenario (run-loop details)

1) user clicks a button w/ onclick
2) that click event is wrapped w/ ember.run (closure action or something lets say)
3) that click event fires this code above
4) when the resolve fires you will most likely not be in that original ember.run (from step 1/2 above) -right?

if you are not as I showed above (total guess -no proof here in code as you can tell) -what is a real use case in which join is successful / useful ?

btw -not trolling here if it sounds that way - just learning a lot about ember.run /when it's used correctly and wanted to learn more :)

taras:
LGTM :+1: to merge

README.md
```diff
@@ -63,12 +64,9 @@ del('/posts/1')                                 // Makes a DELETE request to /po
 
 ### Custom Request Headers
 
-`ember-ajax` allows you to specify headers to be used with a request. This is
-especially helpful when you have a session service that provides an auth token
-that you have to include with the requests to authorize your requests.
+`ember-ajax` allows you to specify headers to be used with a request. This is especially helpful when you have a session service that provides an auth token that you have to include with the requests to authorize your requests.
```
stefanpenner: 80char width please

addon/services/ajax.js
stefanpenner: @toranb runs are intended to be synchronous, so by having an ajax request return asynchronously the run which triggered it should already be complete.

stefanpenner: On another note though, why are we making this change. `Ember.RSVP` promises are exempt from the `auto-run` assertion, to specifically allow for this sort of case, and prevent introducing accidental ordering differences.

stefanpenner: I suspect it is to allow for batching of XHR, i suppose if users are doing that they are in power-user land and must already do the right thing....

taras: @stefanpenner When writing tests that perform async operations, it's useful to be able to easily wait for all promises to resolve before finishing the test. @rwjblue create [`wait()`](https://github.com/switchfly/ember-test-helpers/blob/master/lib/ember-test-helpers/wait.js) for this purpose. This helper relies on the run loop to know if all scheduled timers have been cleared before resolving the `wait()`. Without this, we'd have to track of these promises inside of the service.

stefanpenner: > @rwjblue create wait() for this purpose. This helper relies on the run loop to know if all scheduled timers have been cleared before resolving the wait()

this isn't correct,  `wait` doesn't detect run-lops rather explicitly detects promises async https://github.com/emberjs/ember.js/blob/master/packages/ember-runtime/lib/ext/rsvp.js#L28

taras: I referring to this https://github.com/switchfly/ember-test-helpers/blob/master/lib/ember-test-helpers/wait.js#L37

stefanpenner: Ya, but that is orthogonal, as promises are correctly instrumented to not depend on the legacy polling.

taras: Ok, @alexlafroscia and I will write a few tests. If we're not able to replicate the problem then we'll just close this issue. Thank you for bringing this to my attention.

taras:
@alexlafroscia pls, hold of on merging this. We need to write some tests to see if this is necessary. 

Ref https://github.com/ember-cli/ember-ajax/pull/40#discussion_r49267568

addon/services/ajax.js
stefanpenner: Well, i think we should actually do this, but we should doit for the right reasons. It exists, to correctly enable batching of XHR. as ed does it https://github.com/emberjs/data/blob/master/addon/adapters/rest.js#L857-L859

taras: What does `batching of XHR` provide?

alexlafroscia:
@taras with #42 being pulled in, the code that this was based on has changed a lot.  Is it cleaner to just open a new PR? Or try to rebase this branch?

taras:
It'll likely be easier to just open a new PR, but we need to figure out what to with this first.

alexlafroscia:
Alright, I'll hold off on making that then.

addon/services/ajax.js
rwjblue: @stefanpenner - This is required until I make the wait helper in integration tests handle all promises (as you mentioned the wait helper in acceptance tests already does this). The gain is fairly significant (that we can use ember-ajax from component integration tests), and I don't see a significant downside. 

I'm happy to coordinate be general RSVP awareness in the ember-test-helpers wait helper so that we can remove this wrapping, but in the meantime I think we should land it...

toranb: @stefanpenner @rwjblue thanks for the great rundown (I've learned a ton just reading the comments above). One thing I wanted to clarify quick ... because you are doing a run.join this won't join the original (sync call like a closure action from a button click as you discussed @stefanpenner) but how does this run.join work inside an integration test if I don't wrap my click event (or change event) with ember.run? I ask because as of ember 2.3 I don't wrap my integration clicks/onchanges any longer and I'm curious what the join truly "joins up" with as autorun is disabled in testing

Thank you both in advance -learning 10x from you guys :)

taras:
We're going to go ahead and merge as per https://github.com/ember-cli/ember-ajax/pull/40#discussion_r49268144

We can revisit this in the future it's no longer necessary.

PR #41, opened 2016-01-07
Adds `namespace` option. Fixes #30

### LINKED ISSUE #30: Server API version

manoharank:
cc @taras @alexlafroscia

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,6 +101,23 @@ export default AjaxService.extend({
 });
 ```
 
+### Custom Endpoint Path
+
+The `namespace` property can be used to prefix requests with a specific url namespace.
+
+```js
+// app/services/ajax.js
+
+import Ember from 'ember';
+import AjaxService from 'ember-ajax/services/ajax';
+
+export default AjaxService.extend({
+  namespace: '/api/v1'
+});
+```
+
+`request('/users/me')` would now target `/api/v1/users/me`
+
 ### Custom Host
 
 `ember-ajax` allows you to specify a host to be used with a request. This is
--- a/addon/ajax-request.js
+++ b/addon/ajax-request.js
@@ -22,7 +22,6 @@ const {
   $,
   RSVP: { Promise },
   get,
-  isBlank,
   isPresent,
   run
 } = Ember;
@@ -145,6 +144,7 @@ export default class AjaxRequest {
 
   _buildURL(url, options) {
     const host = options.host || get(this, 'host');
+    const namespace = options.namespace || get(this, 'namespace');
     const urlObject = new RequestURL(url);
 
     // If the URL passed is not relative, return the whole URL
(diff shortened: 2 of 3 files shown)


manoharank:
test-fails

3 test cases are failed due to prepending `/` in the url. Either I have to change the url in the tests or have to add more checks [here](https://github.com/ember-cli/ember-ajax/blob/4ef61fb6d39a8c89adeffc63599cd85021115a34/addon/services/ajax.js#L199).

manoharank:
ping @alexlafroscia @taras

alexlafroscia:
@manoharank I'll check this out.  In the meantime, this PR can no longer be pulled cleanly due to the big changes in #42.  If you could update your PR to pass, that would be excellent.  If not, I would be happy to do so, but would probably close this PR and open a new one based off the current `master` branch.

alexlafroscia:
In the commits since this PR was done, I updated a bunch of those tests so that the URLs make a lot more sense.  If you could update your PR based on this stuff, that would be great.  If you can't, that's fine; I can try to fix whatever is broke, and get it merge-able.  I would like to get this in before the 2.0 release since I think it's a really useful change.

manoharank:
I will update PR this evening

manoharank:
@alexlafroscia Three test cases are failed due to the url (`example.com`) without `http` or leading `/`. How can I fix these existing test cases?

Test case failure for `example.com`. This is due to prepending `/` in `_normalizePath`
test-failure-1
 
Test case failure for `http://example.com`. This is due to `a.href`.
test-failure-2

alexlafroscia:
Huh, weird.  I'll check this out and try to fix whatever is going on with the existing tests.  In the meantime, would you mind adding a few new tests to verify the namespace behavior?

Thanks so much for your work on this!

alexlafroscia:
Okay, I see what's going on: since you added the `_normalizePath` method, and the argument there is being interpreted as part of the path, it's adding the `/` to the beginning of the URL.  I think this behavior is fine, personally, since you're not really meant to put a URL in there that doesn't have the protocol specified.  I'm about to push a commit to this PR that fixed that.

alexlafroscia:
Hmm... I have a commit that I want to add to this PR.  I thought that was a think you can do?  @manoharank if you add me as a collaborator on your fork, that should work, but I thought that I was able to push to your fork being a contributor of the original repo.  @taras any thoughts?

manoharank:
@alexlafroscia I have added you as a collaborator. I will add the test cases for `namespace` option.

alexlafroscia:
Great! I pushed the commit, so you should be able to pull that and add your tests.  Once we have that, this PR gets a :+1: from me!  Thanks again for pushing through to get it finished before the 2.0 release.

manoharank:
@alexlafroscia I have added some test cases and docs, check them and let me know if any changes required.

alexlafroscia:
Looks good! The only thing I'm not sure of is whether we want the namespace overwritable on a per-request basis.  What do you think about that @taras?

taras:
@alexlafroscia no, it should be open option for the service - not over writable.

manoharank:
@alexlafroscia @taras I have removed the overwritable option.

alexlafroscia:
I'm good to pull if you are @taras

taras:
@alexlafroscia looks good to me

PR #42, opened 2016-01-08
Convert project to use ES6 class

### LINKED ISSUE #31: Unclear Upgrade Path for ic-ajax/raw
I have a use case in my application wherein I inspect response codes during long polling:

```javascript
return raw(membershipUrl, { type: 'HEAD', headers: { 'If-Modified-Since': ifModifiedSince } })
  .then((result) => {
    let jqXHR = result.jqXHR;

    // NOTE: if we have already manually refreshed or changed the membership, a different check will
    // already be in progress and we need not make any changes regarding the result of this stale update
    if (this.get('intervalId') !== lastInterval) {
      return;
    }

    if (jqXHR.status === 204) {
      this._newContentAvailable();
    } else {
      this.set('checkInProgress');
    }
  });
```

The upgrade path for such usage is unclear to me, it sort of seems like `¯\_(ツ)_/¯`. In general, I don't feel like headers, and response codes should be totally scrubbed from the response. It'd be nice to opt in to a more advanced use case.

### LINKED ISSUE #35: Remove `ic-ajax` compatibility code
For Ember Ajax 1.0 we should remove already deprecated `ic-ajax` code.

alexlafroscia:
- Converts the AJAX service to us an ES6 class, mixing it into the service using [`ember-mixinify-class`](https://www.npmjs.com/package/ember-mixinify-class) (which came out of the code that was developed for this addon)
- Retains `ic-ajax`-esque `raw` and `request` functions, which call the ES6 class methods "under the hood" -- Closes #31, Closes #35
- Refactors the error-checking methods as functions in their own module, which can also check for instances of their corresponding `AjaxError` classes


taras:
LGTM :+1: to merge

alexlafroscia:
@taras Cool, are we waiting on review from someone else before we pull it in?

taras:
@alexlafroscia no, this one is good to go

PR #43, opened 2016-01-09
Handle server errors the way that Ember Data does

### LINKED ISSUE #22: normalizeErrorResponse should handle JSON responses
My API is returning a JSON response along with an error code status (not any of the ones that are handled), and `normalizeErrorResponse` is converting the payload to a string [here](https://github.com/ember-cli/ember-ajax/blob/master/addon/services/ajax.js#L315). The results in the `detail` of the error object being "[object Object]".

Is this the intended behavior?

alexlafroscia:
Closes #22

### DIFF
--- a/addon/ajax-request.js
+++ b/addon/ajax-request.js
@@ -177,23 +177,24 @@ export default class AjaxRequest {
    */
   handleResponse(status, headers, payload, requestData) {
     payload = payload || {};
+    const errors = this.normalizeErrorResponse(status, headers, payload);
+
     if (this.isSuccess(status, headers, payload)) {
       return payload;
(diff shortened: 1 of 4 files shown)


taras:
I'm not seeing any tests that show detailedMessage being exposed. Was it missed or did I misunderstand how Ember Data code works?

alexlafroscia:
It's now the `message` attribute of the error, which seems to be what Ember Data does, right? And I included a test in the PR that verifies that the message is correct.

alexlafroscia:
Is the attribute supposed to be `detailedMessage`? I just passed the new description as the second parameter to the constructor.

taras:
:+1: LTGM great job!

PR #46, opened 2016-01-19
Fix issue with always sending headers

### LINKED ISSUE #45: Always including headers could lead to unintentional leaking of token credentials
Per https://github.com/ember-cli/ember-ajax#custom-request-headers, if headers are defined they are send on all requests made with `ember-ajax`. The suggested use case here is to send headers including token authentication. 

This seems risky. If anywhere in the application uses `ember-ajax` to make a request to a third-party, these credentials could be leaked. 

Chatted with @taras, he suggested that as a solution headers could only be included by default when using the 'host' option or a relative URL.

alexlafroscia:
This PR ended up being a little bigger than expected; as I dug into the issue referenced in #45, I realized that the behavior of sending a one-off request to a host that didn't match the one specified in the host config options didn't actually do what I assumed it did.  So, some refactoring to be smarter about detecting the host of the request URL took place in addition to only sending the header object when the host matches the specified host, to prevent the headers from leaking through to some other host.

Closes #45

### DIFF
--- a/addon/ajax-request.js
+++ b/addon/ajax-request.js
@@ -29,10 +29,10 @@ const {
 
 export default class AjaxRequest {
 
-  request(url, options) {
-    const hash = this.options(url, options);
+  request(url, options, sendHeaders = false) {
+    const hash = this.options(url, options, sendHeaders);
     return new Promise((resolve, reject) => {
-      this.raw(url, hash)
+      this.raw(url, hash, sendHeaders)
         .then(({ response }) => {
           resolve(response);
         })
@@ -42,8 +42,8 @@ export default class AjaxRequest {
     }, `ember-ajax: ${hash.type} ${hash.url} response`);
   }
 
-  raw(url, options) {
-    const hash = this.options(url, options);
+  raw(url, options, sendHeaders = false) {
+    const hash = this.options(url, options, sendHeaders);
     const requestData = {
       type: hash.type,
       url: hash.url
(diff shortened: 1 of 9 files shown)


taras:
`sendHeaders` property is too heavy handed. If we're not provide an option to not send headers then we should do it through options hash. 

I think we shouldn't have the 3 argument and just use `this. _shouldSendHeaders()` in options method. What do you think?

alexlafroscia:
Yeah, that's fine with me.  I wanted to avoid the need for people to overwrite the hook.  However, if that level of per-request configuration isn't necessary, then removing the extra flag is fine.

taras:
Providing options via arguments is very expensive. It's an API thing - easy to add, very hard to remove. We can always add it later if people find that they do this a lot.

alexlafroscia:
You're totally right.  I'll tweak that and get this PR updated.

alexlafroscia:
@taras what do you think about the behavior shown [in this test](https://github.com/alexlafroscia/ember-ajax/blob/fix-header-sending-issue/tests/unit/ajax-request-test.js#L322-L333)?  It doesn't feel like the right way to handle it, and I'm pretty sure the URL parsing we're doing is powerful enough to detect mismatched hosts even without the protocol being selected.  Should an explicit host _without_ a specified protocol overwrite the default host?

taras:
I'm not sure, this whole: "not specifying host" thing seems very sloppy to me - especially if consider a scenario where the person specified a host. I think the behaviour that you have right now is fine.

alexlafroscia:
Alright, that's fine.  If that part is okay, I think I'm done.

taras:
It looks great good to merge

PR #50, opened 2016-01-30
Allow the wait() helper to work with JSONP

### LINKED ISSUE #48: Doesn't play nice with `wait()` in Integration tests
Just updated to `v2.0.0-beta.1` in an addon project to test whether #28 was fixed - unfortunately the problem persists.

As mentioned there, Testing code (component integration test) like this...
```
// Trigger something that uses `ember-ajax` under the hood
this.$('button').click();

return wait().then(() => {
  // assertions
  asset.ok(...something);
});
```

...won't actually wait for the Ember Ajax request to return.

Will do a bit of investigating now - but just wanted to reopen this.  Thank you~ :heart:

alexlafroscia:
Fixes #48

### DIFF
--- a/addon/ajax-request.js
+++ b/addon/ajax-request.js
@@ -57,6 +57,10 @@ export default class AjaxRequest {
           requestData
         );
 
+        if (hash.dataType === 'jsonp') {
+          $('body').trigger('ajaxComplete');
+        }
+
         if (response instanceof AjaxError) {
           run.join(null, reject, { payload, textStatus, jqXHR, response });
         } else {
@@ -72,8 +76,17 @@ export default class AjaxRequest {
            payload,
            requestData
         );
+
+        if (hash.dataType === 'jsonp') {
+          $('body').trigger('ajaxComplete');
+        }
+
         run.join(null, reject, { payload, textStatus, jqXHR, errorThrown, response });
       };
+
+      if (hash.dataType === 'jsonp') {
+        $('body').trigger('ajaxSend');
+      }
 
       $.ajax(hash);
     }, `ember-ajax: ${hash.type} ${hash.url}`);
--- a/tests/integration/components/jsonp-test.js
+++ b/tests/integration/components/jsonp-test.js
@@ -1,0 +1,36 @@
+import {
+  moduleForComponent,
+  test
+} from 'ember-qunit';
+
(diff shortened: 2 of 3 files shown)


taras:
:+1: to merge

rwjblue:
Hmmm. I'm happy with this as a stop quick BUGFIX, but I would much prefer a solution that doesn't use jQuery to trigger events (this will have issues in FastBoot land for example). 

I think we can handle this and most other scenarios by using a custom test waiter (which is now properly respected in both acceptance tests and ember-test-helpers' wait helper).

alexlafroscia:
Yeah, I wanted to start a conversation with you @rwjblue about to make `wait()` work with this once we start the migration to `fetch`.  I'll look into custom test waiters as the solution to that.  Do you have a link to docs or an example implementation?  I couldn't find anything after a quick search.

rwjblue:
I am basically thinking we do something like:

```js
let pendingRequestCount = 0;
function waitForPendingRequests() {
  return pendingRequestCount === 0;
}

if (Ember.Test) {
  Ember.Test.registerWaiter(waitForPendingRequests);
}

// then in #raw before returning the promise:
pendingRequestCount++;

// in #raw's hash.success callback:
pendingRequestCount--;

// in #raw's hash.error callback:
pendingRequestCount--;
```

alexlafroscia:
Oh yeah, makes sense to me.  Then `ember-ajax` would provide the `waitForPendingRequests` helper, and handle registering it?

rwjblue:
The `Ember.Test.registerWaiter` call sets this up automatically for both acceptance tests and integration/unit tests via the imported `wait()` helper (as of 6 days ago via https://github.com/switchfly/ember-test-helpers/pull/141).

alexlafroscia:
Oh, I see! That's awesome. I'll play around with that a bit and see if I can come up with a custom waiter for `ember-ajax`

hhff:
Thanks guys - is there a release for this change?

alexlafroscia:
@hhff I just dropped a `2.0.0-beta.2`, give that a shot and let me know if that fixes everything for you!

hhff:
Works perfect @alexlafroscia - thanks for your help all!

alexlafroscia:
Glad to hear it!

PR #52, opened 2016-02-02
Ability to  have an array of allowedHosts to set headers for.

webark:
Useful for SOA architectures with different domains communicating across CORS. I could have overwritten the `_shouldSendHeaders` method, but wanted a more strait forward way to tell what domains where whitelisted to pass headers too. Any feedback is much appreciated!

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,10 +101,10 @@ export default AjaxService.extend({
 });
 ```
 
-Headers by default are only passed if the hosts match, or the request is a relitive path.
+Headers by default are only passed if the hosts match, or the request is a relative path.
 You can overwrite this behavior by either passing a host in with the request, setting the
-host for the ajax service, or by setting an array of allowedHosts that can be either
-strings or regular expressions.
+host for the ajax service, or by setting an array of `allowedHosts` that can be either
+an array of strings or regexes.
 
 ```js
 // app/services/ajax.js
--- a/README.md
+++ b/README.md
@@ -103,7 +103,7 @@ export default AjaxService.extend({
 
 Headers by default are only passed if the hosts match, or the request is a relative path.
 You can overwrite this behavior by either passing a host in with the request, setting the
-host for the ajax service, or by setting an array of `allowedHosts` that can be either
+host for the ajax service, or by setting an array of `trustedHosts` that can be either
 an array of strings or regexes.
 
 ```js
@@ -113,7 +113,7 @@ import Ember from 'ember';
 import AjaxService from 'ember-ajax/services/ajax';
 
 export default AjaxService.extend({
-  allowedHosts: [
+  trustedHosts: [
     /\.example\./,
     'foo.bar.com',
   ]
--- a/README.md
+++ b/README.md
@@ -101,6 +101,26 @@ export default AjaxService.extend({
 });
 ```
 
+Headers by default are only passed if the hosts match, or the request is a relitive path.
+You can overwrite this behavior by either passing a host in with the request, setting the
+host for the ajax service, or by setting an array of allowedHosts that can be either
+strings or regular expressions.
+
+```js
+// app/services/ajax.js
+
+import Ember from 'ember';
+import AjaxService from 'ember-ajax/services/ajax';
+
+export default AjaxService.extend({
+  allowedHosts: [
+    /\.example\./,
+    'foo.bar.com',
+  ]
+});
+```
+
+
 ### Custom Endpoint Path
 
 The `namespace` property can be used to prefix requests with a specific url namespace.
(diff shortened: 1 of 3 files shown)


webark:
Hmm.. I look into why the tests re passing locally but not on travis.

webark:
ohhh.. there's no function prototype of bind in the version combo of Node and Phantom that is being used in travis. I'll move it to an arrow function then.

alexlafroscia:
Glad you got that worked out! @taras what do you think about expanding the API to allow this?

taras:
It seems like a natural evolution and a well described use case. I think we should accept it if the code looks good.

alexlafroscia:
Agreed.  I think the code looks good, the PR adds a test and docs, I give it the :shipit:

webark:
I'm a netoriously bad speller.. So let me do a spell check over it, add in a test for a string match, and some error handling and I'll update the PR.

README.md
```diff
@@ -101,6 +101,26 @@ export default AjaxService.extend({
 });
 ```
 
+Headers by default are only passed if the hosts match, or the request is a relitive path.
```
alexlafroscia: Good point on the spelling; `relitive -> relative`

README.md
```diff
@@ -101,6 +101,26 @@ export default AjaxService.extend({
...
+You can overwrite this behavior by either passing a host in with the request, setting the
+host for the ajax service, or by setting an array of allowedHosts that can be either
```
alexlafroscia: If you could put `allowedHosts` in backticks, so it's marked as inline code, I'd appreciate it!

addon/ajax-request.js
```diff
@@ -231,6 +231,24 @@ export default class AjaxRequest {
   }
 
   /**
+   * Match the host to a provided array of host or regex's that can match a host
```
alexlafroscia: `array of strings or regexes` here instead? What do you think?

addon/ajax-request.js
```diff
@@ -231,6 +231,24 @@ export default class AjaxRequest {
...
+   *
+   * @method matchHosts
+   * @public
+   * @param {String} host the host you are sending too
+   * @param {RegExp | String} matcher a string or regex that you can match the host to.
+   * @returns {Boolrsn} if the host passed the matcher
```
alexlafroscia: `Boolean`

alexlafroscia:
@webark I put on my Editor hat and went through the text.  Caught a few spelling things, so I'm really glad you mentioned that!  Fix that up and I'll pull this in.

webark:
@alexlafroscia Thanks for that! I made the changes you suggested, and added in a test that tests against a string as well. I was thinking about raising an error if `allowedHosts` isn't an array, but decided it might be a bit too verbose. Let me know if you think it would be worth while to add that in.

alexlafroscia:
I'm not sure it would really be necessary; if `allowedHosts` is present but not an Array and `.find` is called on it, that will probably throw an error anyway that should make it clear enough that the user needs to go check out the docs :smile:

alexlafroscia:
Oh, actually, one tiny thing I did think of: should we mark `matchHost` as "private" with a `_` in the name of the method?

webark:
@alexlafroscia totally. I was originally going to pass in the whole `RequestURL` object so that one could overwrite that and do some more fancy matching if they wanted to on paths or query params or whatever. I decided that was a bit too verbose though.

webark:
Doh.. Changed the name and forgot to change the reference.

webark:
One more small thing. Do you think that `allowedHosts` name should change? It kind of sounds like the request wouldn't go through if it wasn't in that list, rather then just the headers not go through. Could do something like `hostsForHeaders` or `hosts` or `approvedHosts` or `trustedHosts` or something like that. Just a thought.

alexlafroscia:
That's a really good point too.  I really like `trustedHosts` personally.

webark:
@alexlafroscia I agree. Made that change in code, docs, and tests.

alexlafroscia:
Sweet! I like it. I'd just like @taras to take a peak and then we'll pull it in. Thanks for this!

taras:
It looks great :+1: from me

alexlafroscia:
Thanks again @webark!

webark:
happy to help! I've been sitting on the sidelines of OSS and just being a consumer for a while now.. but have recently decided to join in on the fun! So thanks for making it an enjoyable experience so far!

alexlafroscia:
That's great! This project has some big-ish stuff coming up, that I'm hoping to organize into issues in the near future.  If you like this project help would be awesome!

webark:
Sweet! Looking forward to it. :)

PR #53, opened 2016-02-04
Help users avoid using `.get` for requests

### LINKED ISSUE #51: Warn user if improperly calling `.get()` on AjaxService
For a couple of minutes, I was wondering why this didn't work:

```javascript
this.get('ajax').get('/someResource', { data: /* ... */ });
```

...before realizing that `.get()` is from `Ember.Object`. D:

Do you think it'd be a good idea to perhaps log an error if someone is using that method incorrectly? If so, maybe we could check if the second argument is an object, or if the first arg has a leading slash or something.

alexlafroscia:
Since `.get` might be used for making a `GET` request, but conflicts with Ember's API for getting properties, we can try to detect the incorrect usage and provide the user of the library with direction toward using the correct method.

Closes #51

### DIFF
--- a/addon/ajax-request.js
+++ b/addon/ajax-request.js
@@ -20,6 +20,7 @@ import { RequestURL } from './utils/url-helpers';
 
 const {
   $,
+  Error: EmberError,
   RSVP: { Promise },
   get,
   isPresent,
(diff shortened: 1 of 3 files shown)


taras:
It looks great! :+1:

alexlafroscia:
Excellent 👍

PR #54, opened 2016-02-04
Automatically JSON-encode JSON:API requests

### LINKED ISSUE #49: Content-type and Encoding
Hi there,

I'm trying to send a custom ajax call to my backend server, which is a JSONAPI.org server. I don't know why, but when I try to send a `Content-Type` header (`application/vnd.api+json`) `ember-ajax` is form encoding the request.

```javascript

      this.get('ajax').request('/api/v1/things', {
        method: 'POST',
        data: {
          // ......
        },
        headers: {
          'Content-Type': 'application/vnd.api+json',
          'Accept': 'application/vnd.api+json'
        }
      }).then( jsonapiResponse => {
         // .....
      });
```

Checking my server logs is getting the body sent as a big form encoded blob, rather than JSON.

Am I doing something wrong?

alexlafroscia:
Closes #49

### DIFF
--- a/addon/ajax-request.js
+++ b/addon/ajax-request.js
@@ -26,6 +26,7 @@ const {
   isPresent,
   run
 } = Ember;
+const JSONAPIContentType = 'application/vnd.api+json';
 
 export default class AjaxRequest {
 
@@ -48,6 +49,14 @@ export default class AjaxRequest {
       type: hash.type,
       url: hash.url
     };
+
+    const allHeaders = this._getFullHeadersHash(hash.headers);
+    if (allHeaders['Content-Type'] === JSONAPIContentType) {
+      if (typeof hash.data === 'object') {
+        hash.data = JSON.stringify(hash.data);
+      }
+    }
+
     return new Promise((resolve, reject) => {
       hash.success = (payload, textStatus, jqXHR) => {
         let response = this.handleResponse(
@@ -145,6 +154,17 @@ export default class AjaxRequest {
     options = options || {};
     options.type = method;
     return options;
+  }
+
+  /**
+   * @method _getFullHeadersHash
+   * @private
+   * @param {Object} headers
+   * @return {Object}
+   */
+  _getFullHeadersHash(headers) {
+    const classHeaders = get(this, 'headers') || {};
+    return $.extend(classHeaders, headers);
   }
 
   /**
--- a/addon/ajax-request.js
+++ b/addon/ajax-request.js
@@ -171,7 +171,8 @@ export default class AjaxRequest {
    */
   _getFullHeadersHash(headers) {
     const classHeaders = get(this, 'headers') || {};
-    return merge(classHeaders, headers);
+    const _headers = merge({}, classHeaders);
+    return merge(_headers, headers);
   }
 
   /**
--- a/addon/ajax-request.js
+++ b/addon/ajax-request.js
@@ -24,9 +24,18 @@ const {
   RSVP: { Promise },
   get,
   isPresent,
+  isNone,
+  merge,
   run
 } = Ember;
 const JSONAPIContentType = 'application/vnd.api+json';
+
+function isJSONAPIContentType(header) {
+  if (isNone(header)) {
+    return false;
+  }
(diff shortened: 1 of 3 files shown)


addon/ajax-request.js
```diff
@@ -131,6 +140,17 @@ export default class AjaxRequest {
...
 
   /**
+   * @method _getFullHeadersHash
+   * @private
+   * @param {Object} headers
+   * @return {Object}
+   */
+  _getFullHeadersHash(headers) {
+    const classHeaders = get(this, 'headers') || {};
+    return $.extend(classHeaders, headers);
```
taras: Can you use Ember.merge here?

addon/ajax-request.js
```diff
@@ -47,6 +48,14 @@ export default class AjaxRequest {
       type: hash.type,
       url: hash.url
     };
+
+    const allHeaders = this._getFullHeadersHash(hash.headers);
+    if (allHeaders['Content-Type'] === JSONAPIContentType) {
```
taras: This doesn't allow for JSONAPI requests with an extension. When an extension is used, the extension name is added to the end of the content type. For example: `Content-Type: application/vnd.api+json; ext="ext1,ext2";`

We could do something like
```
function isJSONAPIContentType(header) {
  return header.startsWith('application/vnd.api+json');
}
```

alexlafroscia: Huh, turns out that `.startWith` requires the Babel polyfill.  I didn't realize that.

alexlafroscia:
@taras: I took care of those things, and rebased off of `master` since there was a merge conflict with the tests that I pulled in from the other PR.

Once this gets pulled in, I'm going to go ahead and release Beta 3

addon/ajax-request.js
```diff
@@ -148,6 +164,17 @@ export default class AjaxRequest {
...
 
   /**
+   * @method _getFullHeadersHash
+   * @private
+   * @param {Object} headers
+   * @return {Object}
+   */
+  _getFullHeadersHash(headers) {
+    const classHeaders = get(this, 'headers') || {};
+    return merge(classHeaders, headers);
```
taras: when headers are set on the service, this operation will mutate that headers object. This is likely not the desired behaviour.

You might want to try something like:
```
return {...classHeaders, ...headers};
```

taras:
Other than 1 comment, it looks good.

addon/ajax-request.js
alexlafroscia: Oh, is that how `merge` works? I thought that it just returned a merged version of the two, not mutate the first one and then change it.

Does it rreally mutate the service headers object, though?  `get(this, 'headers') will return an object, and without calling `.set` on the `headers` property, it shouldn't be changed, right?

alexlafroscia: Also, `JSHint` doesn't like the object destructuring approach :frowning:

alexlafroscia: @taras bump :wink:

alexlafroscia:
@taras can you check out my response to the problems you listed?

addon/ajax-request.js
taras: What do you think about this?

```
  /**
   * @method _getFullHeadersHash
   * @private
   * @param {Object} headers
   * @return {Object}
   */
  _getFullHeadersHash(headers) {
    let classHeaders = get(this, 'headers') || {};
    let _headers = merge({}, classHeaders);
    return merge(_headers, headers);
  }
```

alexlafroscia: That's fine too, just not sure why that's necessary over merging `classHeaders` and `headers` directly.  Would mutating `classHeaders` actually change the value of `this.headers`?

taras: Yes, I believe so. You can double check it if you're curious.

alexlafroscia:
So I added the test to check if the original `this.headers` value changes, and the original solution didn't change it.  I still changed the implementation to what you suggested anyway.

taras:
Ok, let's look at it briefly when we pair tomorrow. But this is good to go.

PR #56, opened 2016-02-13
refactor parse-response-headers & add tests

vikram7:
This PR fixes the same issue seen [here](https://github.com/emberjs/data/issues/4122) in Ember data. It also adds test for the `parse-response-headers` util.

### DIFF
--- a/tests/unit/utils/parse-response-headers-test.js
+++ b/tests/unit/utils/parse-response-headers-test.js
@@ -1,0 +1,66 @@
+import parseResponseHeaders from 'ember-ajax/utils/parse-response-headers';
+import { module, test } from 'qunit';
+
+const CRLF = '\u000d\u000a';
+
+module('unit/adapters/parse-response-headers');
+
+test('returns {} when headersString is undefined', function(assert) {
+  let headers = parseResponseHeaders(undefined);
+
+  assert.deepEqual(headers, {}, '{} is returned');
+});
+
+test('header parsing', function(assert) {
+  let headersString = [
+    'Content-Encoding: gzip',
+    'content-type: application/json; charset=utf-8',
+    'date: Fri, 12 Feb 2016 19:21:00 GMT'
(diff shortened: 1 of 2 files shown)


tests/unit/utils/parse-response-headers-test.js
```diff
@@ -0,0 +1,54 @@
...
+  assert.equal(headers['content-type'], 'application/json; charset=utf-8', 'parses header with complex value');
+  assert.equal(headers.date, 'Fri, 12 Feb 2016 19:21:00 GMT', 'parses header with date value');
+});
+
+test('field-name parsing', function(assert) {
+  let headersString = [
+    'value-with-leading-space: value with leading whitespace',
+    'value-without-leading-space:value without leading whitespace',
+    'value-with-colon: value with: a colon',
+    'value-with-trailing-whitespace: banana'
```
alexlafroscia: The key implied that there should be trailing whitespace on the value, but it doesn't look like there is here.  Am I missing something?

alexlafroscia:
Thanks for submitting this! I really appreciate it.

Looking over the Ember Data issue you linked to, there's a few test cases that appear there that don't seem to be here.  Would you mind just reviewing that and making sure we cover all the same cases? One in particular that I noticed is [here](https://github.com/HeroicEric/data/blob/dce22e37735344eef3bd95a46a1196dc48fe2f8c/tests/unit/utils/parse-response-headers-test.js#L29-L39).

tests/unit/utils/parse-response-headers-test.js
vikram7: My mistake! You're right - there should be a trailing whitespace there.

vikram7:
@alexlafroscia Thanks for your comments. I've updated the PR keeping them in mind. Let me know if you think anything else is needed!

alexlafroscia:
Thanks!

PR #62, opened 2016-02-25
Always return payload error status codes as strings

jaswilli:
Error objects' `status` properties are always numeric, except when `AjaxRequest.normalizeErrorResponse` creates and returns its own error object, in which case it coerces the provided `status` into a string.

The result is that error handling code that needs to check status codes has to do `(status === 404 || status === '404')`, etc.

### DIFF
--- a/addon/ajax-request.js
+++ b/addon/ajax-request.js
@@ -505,7 +505,19 @@ export default class AjaxRequest {
    */
   normalizeErrorResponse(status, headers, payload) {
     if (payload && typeof payload === 'object' && payload.errors) {
-      return payload.errors;
+      if (!Ember.isArray(payload.errors)) {
+        return payload.errors;
+      }
+
+      return payload.errors.map(function(error) {
+        let ret = merge({}, error);
+
+        if (typeof ret.status === 'number') {
+          ret.status = `${ret.status}`;
+        }
+
+        return ret;
+      });
     } else {
       return [
         {
--- a/addon/errors.js
+++ b/addon/errors.js
@@ -227,5 +227,6 @@ export function isServerError(error) {
  * @return {Boolean}
  */
 export function isSuccess(status) {
-  return status >= 200 && status < 300 || status === 304;
+  let s = parseInt(status, 10);
+  return s >= 200 && s < 300 || s === 304;
 }
--- a/tests/unit/ajax-request-test.js
+++ b/tests/unit/ajax-request-test.js
@@ -520,6 +520,44 @@ test('it creates a detailed error message for unmatched server errors with a tex
     });
 });
 
+test('it always returns error objects with status codes as strings', function(assert) {
+  assert.expect(1);
+
+  const response = [404, { 'Content-Type': 'application/json' }, ''];
+  server.get('/posts', () => response);
+
+  const service = new AjaxRequest();
+  return service.request('/posts')
+    .then(function() {
+      assert.ok(false, 'success handler should not be called');
+    })
+    .catch(function(result) {
+      assert.strictEqual(result.errors[0].status, '404', 'status must be a string');
+    });
+});
+
+test('it coerces payload error response status codes to strings', function(assert) {
+  assert.expect(2);
+
+  const body = {
+    errors: [
+      { status: 403, message: 'Permission Denied' }
+    ]
+  };
+  const response = [403, { 'Content-Type': 'application/json' }, JSON.stringify(body)];
+  server.get('/posts', () => response);
+
+  const service = new AjaxRequest();
+  return service.request('/posts')
+    .then(function() {
+      assert.ok(false, 'success handler should not be called');
+    })
+    .catch(function(result) {
+      assert.strictEqual(result.errors[0].status, '403', 'status must be a string');
+      assert.strictEqual(result.errors[0].message, 'Permission Denied');
+    });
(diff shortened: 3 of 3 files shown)


alexlafroscia:
This look good to me!  I'd just like @taras to take a look to see if there's a reason we were using strings.

taras:
I'm concerned that this might be a breaking change. Even though it's a bug, apps that rely on current behaviour would break very quietly. 

I would suggest that we change our documentation to encourage the user of error detection helpers. We'd be able to normalize for this in the helpers and provide a smooth upgrade path for 3.0. 

What do you think?

jaswilli:
@taras I think in practice it probably won't end up being a breaking change. There are likely two groups of ember-ajax users who are checking status codes by hand: 1) Those who have yet to be bitten by the fact that ember-ajax generated error objects return statuses as strings, and 2) Those who have, and as a result are necessarily checking for numbers and strings (because passed-through error objects will have numeric statuses). 

 Fixing this now clearly helps those in group 1 and should be non-breaking for those in group 2.

taras:
Actually, we haven't shipped 2.0 yet. So we could include it here with a big WARNING! 

But we should also update the README to tell people to use the error detection helpers.

alexlafroscia:
I think it would be better to use `number`s instead; it's what jQuery itself uses and I think it makes more sense.  Since we're coming up on a breaking change release soon with `2.0`, I think it would be better to change over now and make sure to make a note of it in the release notes.

***

In reference to using the error detection helpers, I'm worried that they're a bit confusing right now, accepting either a status code _or_ the `AjaxError` subclass.  I think generally, we want people to forget about the status code all together and rely on the `AjaxError` subclasses, and should have the helper functions that we provide just detect those instead of doing double duty.

tchak:
From JSON-API spec
> status: the HTTP status code applicable to this problem, expressed as a string value.

Error objects created by `ember-ajax` are intended to be valid JSON-API documents.

jaswilli:
@tchak That's unfortunate. There doesn't even appear to have been any (public) discussion over the decision to pass around numeric status codes as strings. Ugh.

Anyway, this probably isn't the right place to rail against the spec. Assuming that there's a consensus that we want to strictly adhere to the JSON-API error payload specification, should I reverse course in this PR and coerce the upstream `status` from a number into a string?

alexlafroscia:
Hmmm... In light of what @tchak said, I suppose this solution:

> reverse course in this PR and coerce the upstream status from a number into a string

does make more sense.  Maybe that's what we should do, to keep consistent with the JSON-API spec, since that seems to be the direction that the Ember community is moving, given Ember Data's usage of it.  That's what I'm in favor of.

jaswilli:
@alexlafroscia -- Updated to now always coerce numeric status codes in `payload` to strings. The only other new addition is I changed the default `isSuccess` to handle receiving either type of status, since if someone is going to customize it and run `payload.status` through it, they may unknowingly be sending a string and get undesired results.

alexlafroscia:
This looks good to me! It would be a breaking change to the existing behavior though right? Since I know that a lot of people are using the current beta, I wouldn't want to add a breaking change so close to the 2.0 release... @taras I would appreciate your weigh-in on when to pull this in.

jaswilli:
As far as whether or not it's a breaking change, I think it's the same situation as when we were going to do the reverse (https://github.com/ember-cli/ember-ajax/pull/62#issuecomment-189322148).

That said, I'm not personally in any huge hurry to get this merged but I think it's better to get this behavior squared away and consistent sooner rather than later so that when people go to adopt `ember-ajax` they don't run into this issue.

alexlafroscia:
Right; I forgot we had talked about that already. 

I guess we should go ahead and pull this in then, since 2.0 signals a breaking changes anyway, and 3.0 will likely be a long way out. Just want the :+1: from @taras first :smile:

jaswilli:
Why did version 2.0 get released without fixing the `status` handling? 😦

rwjblue:
@jaswilli - I didn't see it when reviewing the 2.0 roadmap issue.  I think we can merge and release as 2.0.1, and yank 2.0.0 (assuming it can be done soon since as of http://blog.npmjs.org/post/141905368000/changes-to-npms-unpublish-policy we can only unpublish within 24 hours).

I do not have `npm` access (looks like only @taras and @alexlafroscia do), could one of you:

1. merge this PR
2. release 2.0.1
3. `npm unpublish ember-ajax@2.0.0`

---

Also, as an aside, the version bump commit (and tag) for 2.0.0 were not pushed to this repositiory.

rwjblue:
> Also, as an aside, the version bump commit (and tag) for 2.0.0 were not pushed to this repositiory.

Turns out the tag was pushed, but the version bump commit wasn't pushed to `master`.  I just pushed it (so that we can land this).

alexlafroscia:
@jaswill @rwjblue thanks for catching that, I will handle that right now.

jaswilli:
Excellent! Thanks @alexlafroscia

alexlafroscia:
Sorry about that again :grimacing: For the future, I'll make sure to rely on Milestone tracking so that these things don't slip through the cracks.

rwjblue:
Ya, it's always tricky for a major bump (good thing we don't do them a ton :smiley_cat😃.

PR #64, opened 2016-02-26
fix spacing regarding JS object syntax

vikram7:
This PR just fixes some JS Object Syntax in the README:

`{key: 'value'}` should read `{ key: 'value' }`


alexlafroscia:
Thanks!

PR #65, opened 2016-02-26
Document error matching helpers

jaswilli:
I took a stab at documenting the error matching helpers, as discussed in #62.

This makes the following changes:

* Adds documentation and examples for the error helpers (e.g., `isNotFoundError`).
* Marks the helper methods on the service as public instead of private (I'm not positive but this seems like it would be desired?)
* Changes the parameter name and type to match those in `errors.js`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,8 +203,7 @@ export default AjaxService.extend({
 
 #### Built in error types
 
-`ember-ajax` has some built in error types that you can use to check error with
-instanceof rather than by comparing to a number. The built in types are:
+`ember-ajax` has built-in error types that will be returned from the service in the event of an error:
 
 * `BadRequestError` (400)
 * `UnauthorizedError`(401)
@@ -215,30 +214,42 @@ instanceof rather than by comparing to a number. The built in types are:
 * `AbortError`
 * `TimeoutError`
 
-Each of these types has a corresponding is* method (like `isBadRequest`) to
-allow you to customize how error is determined.
+#### Error detection helpers
+
+`ember-ajax` comes with helper functions for matching response errors to their respective `ember-ajax` error type. Each of the errors listed above has a corresponding `is*` function (e.g., `isBadRequestError`).
+
+Use of these functions is **strongly encouraged** to help eliminate the need for boilerplate error detection code.
 
 ```js
 import Ember from 'ember';
-import {UnauthorizedError, ForbiddenError} from 'ember-ajax/errors';
+import {isNotFoundError, isForbiddenError} from 'ember-ajax/errors';
 
 export default Ember.Route.extend({
   ajax: Ember.inject.service(),
   model() {
-    return this.get('ajax').request('/user/me')
-      .catch(function(error){
-        if (error instanceof UnauthorizedError) {
-          // user is not logged in
-          this.transitionTo('login');
-        } else if (error instanceof ForbiddenError) {
-          // user doesn't have access to
-          this.transitionTo('index');
+    const ajax = this.get('ajax');
+
+    return ajax.request('/user/doesnotexist')
+      .catch(function(error) {
+        if (isNotFoundError(error)) {
+          // handle 404 errors here
+          return;
         }
+
+        if (isForbiddenError(error)) {
+          // handle 403 errors here
+          return;
+        }
+
+        // other errors are handled elsewhere
         throw error;
       });
   }
 });
 ```
+
+If your errors aren't standard, the helper function for that error type can be used as the base to build your custom detection function.
+
 
 ## Testing
 


README.md
```diff
@@ -240,6 +238,63 @@ export default Ember.Route.extend({
...
+
+export default Ember.Route.extend({
+  ajax: Ember.inject.service(),
+  model() {
+    const ajax = this.get('ajax');
+    const {isNotFoundError} = ajax;
+
+    return ajax.request('/user/doesnotexist')
+      .catch(function(error) {
+        if (ajax.isNotFoundError(error)) {
```
alexlafroscia: I'm not sure that this is actually how these were intended to be used...

In my mind, at leas, the `isNotFoundError` methods were private on the service so that they could be overwritten as needed for usage within the `ajax` service itself.  If you want to do error detection within the application, you import the method from `ember-cli-ajax/errors` and use it directly.

taras:
@alexlafroscia is right, the methods on the class covert status codes into exceptions. The helpers in `ember-cli-ajax/errors` are the helpers to be used for detection.

jaswilli:
My thought process was that if we want people to be able to easily use the helper methods and to be able to easily customize them should it be needed, that dealing with the methods that live on the `AjaxService` would be much more convenient (and therefore more likely to be used). I also thought it would be a good thing if the `AjaxService` methods took advantage of any custom error detection logic, so you'd always end up with the "right" type of `AjaxError`.

At least as far as the customization goes, it's much more cumbersome to have to wrap the error helpers in a new file, i.e.:

```js
// app/utils/ajax-error-helpers.js

import {
  isNotFoundError as _isNotFoundError,
  isBadRequestError as _isBadRequestError
 // ...
} from 'ember-ajax/errors';

export function isNotFoundError(error) {
  if (_isNotFoundError(error)) {
    return true;
 }}

  // custom whatever
}
```

Anyway, if you feel strongly that the `AjaxService` methods should be private and all error matching should be done directly with the exported `errors.js` functions I'm happy to re-work the write-up.

alexlafroscia:
I totally see what you're saying, but in the example you provided you're invoking the custom handler logic twice when you don't need to be.

In the example you provided in the `README`, you're using the `ajax.isNotFoundError` method to detect if the error thrown from `ajax.request` was a `NotFoundError`, but if you've overwritten that method and provided your own implementation, then that handler has already run once and determined the correct type for the `error` argument to the `.catch` handler.  Because of that, you can use the provided helper functions from `ember-ajax/errors` to detect if the `error` argument is a `NotFoundError`; you don't need to invoke the custom logic a second time.

To me, the notes about overriding `isNotFoundError` are spot on, but what needs to change is the fact that the `.catch` handler uses the `ajax.isNotFoundError` handler a second time, since that code has already run once and generated the `NotFoundError` in the first place.

***

Do you think it would be easier to differentiate between the two cases, being:

1. The service's `is_____Error` methods, which are used to detect the conditions for a `______Error`
2. The helper functions, which are for detecting the error types themselves

if they were named differently?  I also feel like it might be confusing that the helper functions can take either an `int` or a subclass of `AjaxError`; this was done for convenience, but in hindsight is probably confusing since it makes their intended usage harder to determine.

jaswilli:
I updated the PR to document the helper functions from `errors.js` instead of the service methods.


> Do you think it would be easier to differentiate between the two cases, being:
> 
> The service's is_____Error methods, which are used to detect the conditions for a ______Error
> The helper functions, which are for detecting the error types themselves
> if they were named differently?

It seemed natural to me to use the service's `is*` methods because I was going to customize them to handle the non-standard error payloads I was dealing with, and it's way more convenient to override a method on an ember object than it is to compose new helper functions out of `errors.js`.

> I also feel like it might be confusing that the helper functions can take either an int or a subclass of AjaxError; this was done for convenience, but in hindsight is probably confusing since it makes their intended usage harder to determine.

It's maybe a tiny bit confusing, but on the other hand, it's also kind of nice to be like "pass in whatever you have and if I can match it against an error you're good..."

alexlafroscia:
> It's maybe a tiny bit confusing, but on the other hand, it's also kind of nice to be like "pass in whatever you have and if I can match it against an error you're good..."

This is the spirit with which they were designed, so I'm glad that that's not confusing.

> It seemed natural to me to use the service's is* methods because I was going to customize them to handle the non-standard error payloads I was dealing with, and it's way more convenient to override a method on an ember object than it is to compose new helper functions out of errors.js.

You had the right intuition, and that's exactly what we want users to do!  The only issue is that when it comes to detecting the error within the `.catch` method's callback, you don't have to use the custom error detector again since the `.request` method on the service has already run the custom one.

README.md
```diff
@@ -226,7 +224,7 @@ export default Ember.Route.extend({
   ajax: Ember.inject.service(),
   model() {
     return this.get('ajax').request('/user/me')
-      .catch(function(error){
+      .catch(function(error) {
         if (error instanceof UnauthorizedError) {
```
taras: This should use an error detection helper.

jaswilli: I left that code snippet alone because it was providing an example of the `AjaxError` subtypes.

Perhaps the that entire code block should be removed to focus on the one below that uses the helper function?

taras: Good call. Let's remove the code example and just leave the list of types.

jaswilli:
Updated to remove the code snippet illustrating the use of checking the error type via `instanceof`.

alexlafroscia:
👍 looks good to me!

jaswilli:
Are we still happy with these doc changes? Since all the error detection helpers shipped with 2.0 it would be nice to encourage people to rely on them instead of `instanceof` against the `AjaxErrors` themselves.

alexlafroscia:
Yeah, I think so.  I was waiting for @taras to respond one way or another, but I think I'm okay with this.

Issue #70, opened 2016-03-07
Problematic deletion of trailing slashes in urls


fpauser:
See https://github.com/ember-cli/ember-ajax/blob/master/addon/ajax-request.js#L236

Why are trailing slashes removed?
Right now to keep the trailing slashes in pathes we must convert paths to absolute urls, which feels just not right.

alexlafroscia:
Currently, that's just done for convenience on our part.  You have an API that requires a slash at the end?

badazz91:
Our custom API reacts somewhat opinionated, when forgetting the trailing slashes, resolving in a 404 error.

topaxi:
I just ran into this, I'm encountering an API which does a redirect from non-slashed endpoint to an endpoint with an slash and messes up the response.

alexlafroscia:
Sorry about that; there's a PR open that needs some tweaking before we can fix this. I will try to clean it up myself this week.

kaungst:
One workaround I found was to add an additional trailing slash at the end.
e.g. http://www.random.com/stuff//

jamesarosen:
For posterity: we just fixed this in our app by upgrading to 2.4.1 :)

alexlafroscia:
@jamesarosen excellent! Glad to hear it's working for oyu

PR #71, opened 2016-03-08
Do not remove trailing slashes, when dealing with relative urls.

badazz91:
see #70

### DIFF
--- a/addon/ajax-request.js
+++ b/addon/ajax-request.js
@@ -231,11 +231,6 @@ export default class AjaxRequest {
       if (path.charAt(0) !== '/') {
         path = `/${path}`;
       }
-
-      // remove end `/`
-      if (path.charAt(path.length - 1) === '/') {
-        path = path.slice(0, -1);
-      }
     }
     return path;
   }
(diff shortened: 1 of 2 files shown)


alexlafroscia:
@badazz91 can you add me as a collaborator on your fork? I added a failing test that is introduced by your changes.

badazz91:
@alexlafroscia I added you as a collaborator.

alexlafroscia:
@badazz91 great, thanks.  I added the commit with the failing test.

badazz91:
@alexlafroscia We made some modifications, to satisfy the failing test. Let me know about your thoughts.

alexlafroscia:
I added some more test cases, found a place where they fail again.  I still feel pretty strongly that `_normalizePath` should be left as it was, to ensure that all the segments end up in the same format (`/` at the start, no `/` at the end).  Maybe checking if the full URL has a trailing slash, and then add it back at the end if it needs it?

alexlafroscia:
Closed in favor of #96 which was merged a while back.

PR #75, opened 2016-03-11
Basic fastboot support

topaxi:
This adds basic fastboot support by using the provided `URL` module by ember-fastboot-server and [`najax`](https://github.com/najaxjs/najax).

I get warnings as `najax` does not implement `jqXHR.getAllResponseHeaders()`, but this should work for basic ajax requests.
It seems to be implemented in `najax` on the `master` (https://github.com/najaxjs/najax/pull/26) branch, I'm not sure if this is released in a npm package yet.

### DIFF
--- a/addon/ajax-request.js
+++ b/addon/ajax-request.js
@@ -19,6 +19,7 @@ import {
 } from './errors';
 import parseResponseHeaders from './utils/parse-response-headers';
 import { RequestURL } from './utils/url-helpers';
+import ajax from './utils/ajax';
 
 const {
   $,
@@ -123,7 +124,7 @@ export default class AjaxRequest {
 
       this.pendingRequestCount++;
 
-      $.ajax(hash);
+      ajax(hash);
     }, `ember-ajax: ${hash.type} ${hash.url}`);
   }
 
@@ -489,7 +490,7 @@ export default class AjaxRequest {
     let json = responseText;
 
     try {
-      json = Ember.$.parseJSON(responseText);
+      json = $.parseJSON(responseText);
     } catch (e) {}
 
     return json;
--- a/addon/utils/ajax.js
+++ b/addon/utils/ajax.js
@@ -1,0 +1,9 @@
+/* global najax */
+import Ember from 'ember';
+import isFastBoot from './is-fastboot';
+
+const {
+  $
+} = Ember;
+
+export default isFastBoot ? najax : $.ajax;
--- a/addon/utils/is-fastboot.js
+++ b/addon/utils/is-fastboot.js
@@ -1,0 +1,3 @@
+/* global FastBoot */
+const isFastBoot = typeof FastBoot !== 'undefined';
+export default isFastBoot;
--- a/addon/utils/url-helpers.js
+++ b/addon/utils/url-helpers.js
@@ -1,4 +1,5 @@
-/* global require, module */
+/* global require, module, URL */
+import isFastBoot from './is-fastboot';
 
 const absoluteUrlRegex = /^(http|https)/;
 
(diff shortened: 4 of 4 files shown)


alexlafroscia:
Would we need to include `najax` as a dependency for this to work? Or would it be included some other way?

topaxi:
`najax` is provided to the sandbox by `ember-fastboot-server`: [ember-fastboot-server/lib/ember-app.js#L32](https://github.com/ember-fastboot/ember-fastboot-server/blob/master/lib/ember-app.js#L32)

alexlafroscia:
Awesome, that's good to know.  My one concern is accepting this without automated testing in Fastboot... What do you think @taras?

topaxi:
Yeah, tests would be great, I don't know the current state on fastboot testing.

Related discussions: https://github.com/tildeio/ember-cli-fastboot/issues/111 and https://github.com/tildeio/ember-cli-fastboot/issues/68

alexlafroscia:
Thanks for linking those, I was just reading them myself.  Maybe we should pull it in, add _real_ tests once it's supported, and update the README to say that Fastboot is supported... Unofficially?

alexlafroscia:
I made a little example app, and I think that the lack of support for `getAllResponseHeaders` is going to be a deal breaker for now.  I tried making a little example app that just hits Github's API and grabs the list of issue for this repo, and immediately got an error due to `najax.getAllResponseHeaders` not being implemented.

alexlafroscia:
Also, one other thing I would like to do: if we can't `import` NAJAX as a module, could we expose our own module that exports `najax` and consume it within the addon that way?  Either that, or add `najax` to the globals list in the `.jshintrc`, I don't love declaring it at the top of the file.

topaxi:
`getAllResponseHeaders` is implemented in the latest `najax` release and also in the latest `ember-fastboot-server` (though the latest `ember-fastboot-server` isn't released yet`.

It works for me as long as I don't need any headers (the `getAllResponseHeaders` error is just a warning in my case), when the next `ember-cli-fastboot` is released, this should work too.

I'll look into creating an ajax module which exports either `najax` or `jQuery.ajax`.

alexlafroscia:
> I'll look into creating an ajax module which exports either najax or jQuery.ajax.

That sounds perfect

> getAllResponseHeaders is implemented in the latest najax release and also in the latest ember-fastboot-server (though the latest ember-fastboot-server isn't released yet`.

I see, I'll look see if I can get it working with the current `master` then.

alexlafroscia:
I think the issue that I'm seeing is that the `najax` version that comes bundled with Fastboot right now fails when making the request that I'm trying to make.  I tried it using the latest release of `najax`, and [the request seems to work fine](https://tonicdev.com/56e86e4e0cd3711100aa9840/56e86e4e0cd3711100aa9841).

Assuming that this works correctly (and as far as I can tell, it does), I'm in favor of pulling this in and gaining basic Fastboot support once those few changes are made regarding importing `ajax` from a separate module.

alexlafroscia:
Huh, I keep getting the following error with `najax`:

> Ember Data Request GET https://api.github.com/repos/ember-cli/ember-ajax/issues?state=open returned a 0\nPayload (Empty Content-Type)\n"SyntaxError: Unexpected token R\\n    at Object.parse (native)\\n    at IncomingMessage.<anonymous> (/Users/alex/Projects/open-source/tmp/fastboot-test/node_modules/ember-cli-fastboot/node_modules/ember-fastboot-server/node_modules/najax/lib/najax.js:135:23)\\n    at emitNone (events.js:85:20)\\n    at IncomingMessage.emit (events.js:179:7)\\n    at endReadableNT (_stream_readable.js:906:12)\\n    at nextTickCallbackWith2Args (node.js:474:9)\\n    at process._tickCallback (node.js:388:17)"

Doesn't seem to happen using either `najax@0.4.0` or `0.3.2` when using `najax` by itself, it only happens in Fastboot.  Request works fine in the browser, too.

topaxi:
Rebased and moved the `ajax` function into a module in `utils`. I also moved `is-fastboot` into `utils` as I think it fits better in there.

tchak:
I started some work on using `ember-network` integration. This will give us a first class fastboot support. See #81.

alexlafroscia:
Sorry I haven't gotten to check on this yet.  Been traveling all weekend, I'll get to this today.

alexlafroscia:
Okay, so now that `2.0` has landed, I think that we should pull this into `master` so that people that want this have a way of getting it.

PR #77, opened 2016-03-15
Implement custom waiter

alexlafroscia:
As suggested here: https://github.com/ember-cli/ember-ajax/pull/50#issuecomment-177342649

Probably required for Fastboot support, since we don't want to rely on `$('body').trigger('ajaxComplete')` to make the `wait()` helper work correctly.


alexlafroscia:
@rwjblue could you make sure that this was done the way you were envisioning?

Issue #76, opened 2016-03-14
Upload with feedback


mbrakken:
Apologies if this is clarified somewhere, but I've only had time to take a cursory glance at this project and related specs.

I understand that the fetch API will not have any upload progress callbacks since it'll use streams, and that the xhr object used to polyfill future behavior isn't exposed since it won't exist with fetch. I also assume, though haven't verified, that streams are currently not widely supported generally, and are not supported here. 

I have an app that supports upload of large files (eg. wav pcm, flac, etc), and it's important that I can provide feedback on the state of an upload operation.

Currently using ic-ajax, I have a service that defines an xhr object so I can set callbacks on relevant event listeners:
```javascript
_upload(file, url, headers, requestType, callbacks){
    return ajax({
        xhr: function(){
            var xhr = new window.XMLHttpRequest();
            xhr.upload.addEventListener('progress', function(evt){
                if (evt.lengthComputable) {
                    var progress =  Math.round((evt.loaded / evt.total)*100);
                    callbacks.upload.progress(progress);
                }
            }, false);
            xhr.upload.addEventListener('load', function(evt){
                callbacks.upload.load(evt);
            }, false);
            // etc
        },
        contentType : fileType,
        url         : url,
        type        : requestType,
        headers     : headers,
        data        : file,
        processData : false,
        dataType    : 'json'
    });
}
```
Since ember ajax initially provided swap out support for ic-ajax, does it offer some means to support similar behavior, or do I need to switch to using `$.ajax` for upload if I want to migrate to ember ajax?

alexlafroscia:
You _should_ be able to do the same thing using `ember-ajax/request` -- that is provided to be a drop-in replacement for `ic-ajax`.  Did you give that a shot?

alexlafroscia:
I'm going to close this for now, since uploading with feedback isn't something we really intend to support.  I wish I had a better answer but we've decided to keep the API as `fetch`-compliant as we can, since the long-term goals are to leverage that instead of jQuery.

If uploading with feedback is required, I'd suggest just using `$.ajax` directly for that request.

PR #81, opened 2016-04-02
[WIP] use ember-network fetch

tchak:
This is a WIP on using `ember-network` fetch as low level interface for `ember-ajax` as discussed with @taras and @tomdale at EmberConf.

First problem. `ember-network` is broken when required from an addon. @tomdale could you have a look?

```
⠋ Buildingapp.import is not a function
TypeError: app.import is not a function
    at Class.module.exports.included (/Users/tchak/git/github/ember-cli/ember-ajax/node_modules/ember-network/index.js:17:15)
    at /Users/tchak/git/github/ember-cli/ember-ajax/node_modules/ember-cli/lib/models/addon.js:244:32
```

Other questions I have so far:
- `fetch` have no timeout support. Should we implement one? We will not be able to abort the request but at least it will not block the promise forever.
- We need to decide on how we want to flag fetch support. `ENV.USE_FETCH`?

### DIFF
--- a/addon/ajax-request.js
+++ b/addon/ajax-request.js
@@ -1,4 +1,6 @@
 import Ember from 'ember';
+import fetch from "ember-network/fetch";
+
 import {
   AjaxError,
   UnauthorizedError,
@@ -63,15 +65,12 @@ export default class AjaxRequest {
         .catch(({ response }) => {
           reject(response);
         });
-    }, `ember-ajax: ${hash.type} ${hash.url} response`);
+    }, `ember-ajax: ${hash.method} ${hash.url} response`);
   }
 
   raw(url, options) {
-    const hash = this.options(url, options);
-    const requestData = {
-      type: hash.type,
-      url: hash.url
-    };
+    let hash = this.options(url, options);
+    let { method } = hash;
 
     if (isJSONAPIContentType(hash.headers['Content-Type'])) {
       if (typeof hash.data === 'object') {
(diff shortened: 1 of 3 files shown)


alexlafroscia:
Just as a warning: I think this is going to be a pretty major, breaking change, warranting a 3.0 release.  As far as I know, `$.ajax` and the `fetch` API aren't exactly equivalent, in the fact that there are certain `options` that can be passed to jQuery AJAX that `fetch` does not have an analogue for.  Moving to `ember-networking` under the hood _is_ something that we want to do in the long term, though, so I'm really happy that you're starting to take a look at what that takes!

stefanpenner:
https://github.com/tomdale/ember-network/pull/4 will be a pre-req for this being merged.

locks:
@stefanpenner #4 is merged ;P

alexlafroscia:
@locks the hangup here is the fact that I don't think `fetch` can emulate everything that `$.ajax` can do, at least not out of the box.  As far as I know, no one has an implementation of `$.ajax` that supports all the same options but uses `fetch` under the hood, so there's a considerable amount of work to do to get that working.

btecu:
@alexlafroscia What are the features that `ajax` can do but `fetch` can't?

tomdale:
@btecu Cancelation is the big one, but `fetch()` also doesn't have native XML parsing AFAIK.

alexlafroscia:
I've been thinking about this lately and... I'm not sure if we should actually ever do this.

With the availability of [`ember-network`](https://github.com/tomdale/ember-network) and [`ember-fetch`](https://github.com/stefanpenner/ember-fetch) (which both seem to scratch the itch of needing a Fastboot/Browser-compatible `fetch` implementation) I don't really see why one would want to use `fetch` within `ember-ajax` as well.

Presumably jQuery-style AJAX isn't going anywhere anytime soon, and knowing that jQuery might eventually be dropped as an Ember dependency gives an even better reason for keeping this addon `$.ajax`-specific, since we could presumably pull in _just_ the networking part of jQuery as a dependency of this addon so that Ember can drop jQuery without everyone else needing to deal with that.

I'm definitely open to thoughts and insights from everyone else.  I think, however, if we decide that `fetch` is no longer part of the roadmap, we can start thinking about useful features that would not necessarily be `fetch`-compatibly (like monitoring progress, which was requested by denied in #76)

alexlafroscia:
I'm going to close this PR for now; we're planning to open an RFC soon to discuss switching to `fetch` instead. We can always re-open it later if that makes sense.

PR #84, opened 2016-04-08
Add a `delete` alias

sandstrom:
Since the other methods map to the HTTP type name (PATCH => `patch()`, POST => `post()`) it makes sense to do the same for `delete`. Does this sound like a good idea?

### DIFF
--- a/addon/ajax-request.js
+++ b/addon/ajax-request.js
@@ -164,8 +164,8 @@ export default class AjaxRequest {
    * alias for `del()`
    * @public
    */
-  delete(url, options) {
-    return this.del(url, options);
+  delete(...args) {
+    return this.del(...args);
   }
 
   /**
--- a/addon/ajax-request.js
+++ b/addon/ajax-request.js
@@ -164,8 +164,8 @@ export default class AjaxRequest {
    * alias for `del()`
    * @public
    */
-  delete(...args) {
-    return this.del(...args);
+  delete() {
+    return this.del(...arguments);
   }
 
   /**
--- a/addon/ajax-request.js
+++ b/addon/ajax-request.js
@@ -160,6 +160,15 @@ export default class AjaxRequest {
   }
 
   /**
+   * calls `request()` but forces `options.type` to `DELETE`
+   * alias for `del()`
+   * @public
+   */
+  delete(url, options) {
+    return this.del(url, options);
+  }
+
+  /**
    * Wrap the `.get` method so that we issue a warning if
    *
    * Since `.get` is both an AJAX pattern _and_ an Ember pattern, we want to try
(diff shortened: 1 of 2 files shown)


alexlafroscia:
We [already have `del`](https://github.com/ember-cli/ember-ajax/blob/ec3d68da8644d7eab6a65e5fd165a6208cd9fba2/addon/ajax-request.js#L154-L160), and I think that's probably enough.  Does that seem alright to you?

sandstrom:
@alexlafroscia I know (it calls `del` internally), but the others match the respective http method. Not a big issue in any way, but thought it would be neat to keep them aligned where possible.

Reducing the surprises by providing both `del` and `delete` will make the addon more pleasant to use.

alexlafroscia:
I'm fine with including both, personally.  This would just need two things then:

1. Call `this.del` using `...arguments` instead of passing them explicitly
2. Add tests to ensure that this works correctly

sandstrom:
@alexlafroscia Good suggestions!

I've made both adjustments (tests are still running but should work). There are three commits, but with the recent [*Squash and Merge*](https://github.com/blog/2141-squash-your-commits) I hope that's alright.

alexlafroscia:
Great! Thanks for that.  I just have one tiny tweak; I'll comment on the line.  With the Squash and Merge don't worry about adding another commit, I'll squash from my end :thumbsup:

addon/ajax-request.js
```diff
@@ -160,6 +160,15 @@ export default class AjaxRequest {
   }
 
   /**
+   * calls `request()` but forces `options.type` to `DELETE`
+   * alias for `del()`
+   * @public
+   */
+  delete(...args) {
```
alexlafroscia: We don't actually need this here; you can remove `...args` from here...

addon/ajax-request.js
```diff
@@ -160,6 +160,15 @@ export default class AjaxRequest {
...
+    return this.del(...args);
```
alexlafroscia: ... and change this to `...arguments` instead, since `arguments` will always be defined as an array-like object of all arguments passed to the function.

sandstrom: The code above basically does the same thing.

I can change it if you really want to, but it won't much of a difference.

```js
# try in FF console
let a = function(...args) { console.log(...args); args.forEach(function(a) { console.log(`a: ${a}`); }); };
a('bat', 'bird', 'badger');

let b = function(...arguments) { console.log(...arguments); arguments.forEach(function(a) { console.log(`b: ${a}`); }); };
b('bat', 'bird', 'badger');
```

jaswilli: @sandstrom 

You can apply the spread operator to the function's implicit `arguments` variable, so there's no need to put it in the function signature. I think that's what @alexlafroscia is getting at. 😄 
```
delete() {
  return this.del(...arguments);
}
```

alexlafroscia: @jaswilli exactly, that's what I meant.  Sorry that I didn't make that clear enough.

sandstrom:
@alexlafroscia done

PR #88, opened 2016-04-09
Update description

zoltan-nz:
because of emberobserver.com: https://www.emberobserver.com/addons/ember-ajax ;)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "ember-ajax",
   "version": "2.0.0-beta.3",
-  "description": "The default blueprint for ember-cli addons.",
+  "description": "Service for making AJAX requests in Ember 1.13+ applications.",
   "directories": {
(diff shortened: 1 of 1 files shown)


alexlafroscia:
Do you know if there's some way to signal that 1.12 isn't supported anymore, other than doing this?  I feel like the `package.json` description is the wrong place to note this, and that maybe the README would be more appropriate.

zoltan-nz:
EmberObserver uses package.json description line as the default description on the home page. It looks, that package.json description was never updated from default, you can see this on the home page:

image

I just used the first sentence from README.md, it is mainly describe what is this `ember-ajax` addon for. ;)

zoltan-nz:
npmjs.org uses description as subtitle also.

image

alexlafroscia:
Ahh, I see.  This is :thumbsup: from me then!

PR #92, opened 2016-04-11
isAjaxError helper

### LINKED ISSUE #90: Add isAjaxError helper?
I'm updating some projects to use the new `is*` error helper functions and there are some cases where I don't care at all what specific `AjaxError` came back, just knowing that it's an `AjaxError` is enough.

For consistency, does it make sense to add an `isAjaxError()` helper as well? Otherwise it will be necessary to do a mix of `instanceof` checks and `is*` checks. It seems cleaner if everything could be handled with the helpers.

jaswilli:
Closes #90 

* Adds an `isAjaxError` helper function to match the other `is*` error helpers.
* Updates the README to mention `AjaxError` and `isAjaxError`.
* Adds test coverage for `NotFoundError`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -171,12 +171,6 @@ You can even leave off the forward slash if you'd like:
 request('users/me')
 ```
 
-### Error handling
-
-`ember-ajax` provides built in error classes that you can use to check the error
-that was returned by the response. This allows you to restrict determination of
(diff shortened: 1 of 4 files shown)


alexlafroscia:
Thank you!

PR #93, opened 2016-04-12
Use `Ember.assign()` if possible

Cryrivers:
As `Ember.merge` is deprecated since `Ember 2.5.0`

### DIFF
--- a/addon/ajax-request.js
+++ b/addon/ajax-request.js
@@ -27,11 +27,11 @@ const {
   RSVP: { Promise },
   get,
   isNone,
-  merge,
   run,
   Test,
   testing
 } = Ember;
+const assign = Ember.assign || Ember.merge;
 const JSONAPIContentType = 'application/vnd.api+json';
 
 function isJSONAPIContentType(header) {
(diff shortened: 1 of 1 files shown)


addon/ajax-request.js
```diff
@@ -200,8 +200,8 @@ export default class AjaxRequest {
    */
   _getFullHeadersHash(headers) {
     const classHeaders = get(this, 'headers') || {};
-    const _headers = merge({}, classHeaders);
-    return merge(_headers, headers);
+    const _headers = assign({}, classHeaders);
```
rwjblue: `assign` allows more than two arguments so we should be able to do this in a single invocation:

```js
return assign({}, classHeaders, headers);
```

rwjblue: Ahh, but we can't do that until we can require `v2.5.0`, :crying_cat_face:

rwjblue:
This LGTM

addon/ajax-request.js
Cryrivers: @rwjblue yep, `assign` could be `Ember.merge`. :trollface:

alexlafroscia:
Hmm, @cryrivers CI doesn't seem very happy with the change, but it looks good to me once that is settled.

Cryrivers:
@alexlafroscia not sure why those 5 test cases failed :cry:

orf:
Looks like the deprecation has been removed: https://github.com/emberjs/ember.js/pull/13344

alexlafroscia:
Yeah, in light of that fact, I'm going to close this.  Thanks for the PR though, @Cryrivers!

PR #96, opened 2016-04-23
Prevent removing trailing slash in request URL

alexlafroscia:
The old way of building a URL was too brittle and didn't cover a lot of the different cases that could take place, in different combinations of the host, namespace, and segment having or not having slashes included.

The new approach implemented here handles the various cases correctly, and ensures that we do not remove a trailing slash if one was provided.

### DIFF
--- a/addon/mixins/ajax-request.js
+++ b/addon/mixins/ajax-request.js
@@ -43,6 +43,27 @@ function isJSONAPIContentType(header) {
   return header.indexOf(JSONAPIContentType) === 0;
 }
 
+function startsWithSlash(string) {
+  return string.charAt(0) === '/';
+}
+
+function endsWithSlash(string) {
+  return string.charAt(string.length - 1) === '/';
+}
+
+function stripSlashes(path) {
+  // make sure path starts with `/`
+  if (startsWithSlash(path)) {
+    path = path.substring(1);
+  }
+
+  // remove end `/`
+  if (endsWithSlash(path)) {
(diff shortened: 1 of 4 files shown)


addon/ajax-request.js
```diff
@@ -42,6 +42,35 @@ function isJSONAPIContentType(header) {
...
+
+function endsWithSlash(string) {
+  if (typeof string === 'string') {
+    return string.charAt(string.length - 1) === '/';
+  }
+  return string;
+}
+
+function stripSlashes(path) {
+  if (typeof path === 'string') {
```
taras: Instead of checking for string 3 times, you could either type cast the argument to a string or throw an error if the argument it not a string.

taras:
Other than one comment, it looks good.

addon/ajax-request.js
alexlafroscia: > throw an error if the argument it not a string
Which argument? The one to `_buildURL`?

taras: No, I'm referring to `typeof path === 'string'`. You don't need to check the argument type here.

It's safe enough for stripSlashes to either force `path` argument to be a string or throw an error here if it's not.

alexlafroscia: Oh okay, gotcha, I'll fix that up.

alexlafroscia:
@taras can you review this again? I took care of the issue you mentioned.

taras:
👍

PR #105, opened 2016-05-04
Fix memory leak in tests

### LINKED ISSUE #104: Memory Leak in Tests
I'm converting an Ember 1.13 app from ember-cli-ic-ajax to ember-ajax and my tests keep running out of memory. See the discussion below for the (likely) cause.

alexlafroscia:
Closes #104

### DIFF
--- a/addon/ajax-request.js
+++ b/addon/ajax-request.js
@@ -52,7 +52,9 @@ export default class AjaxRequest {
 
   init() {
     if (testing) {
-      Test.registerWaiter(() => (pendingRequestCount === 0));
+      Test.registerWaiter(function() {
+        return pendingRequestCount === 0;
+      });
     }
   }
 
--- a/addon/ajax-request.js
+++ b/addon/ajax-request.js
@@ -43,20 +43,13 @@ function isJSONAPIContentType(header) {
 }
 
 let pendingRequestCount = 0;
+if (testing) {
+  Test.registerWaiter(function() {
+    return pendingRequestCount === 0;
+  });
+}
 
 export default class AjaxRequest {
-
-  constructor() {
-    this.init();
-  }
-
-  init() {
-    if (testing) {
-      Test.registerWaiter(function() {
-        return pendingRequestCount === 0;
-      });
-    }
-  }
 
   request(url, options) {
     const hash = this.options(url, options);
--- a/addon/ajax-request.js
+++ b/addon/ajax-request.js
@@ -216,7 +216,6 @@ export default class AjaxRequest {
     options.url = this._buildURL(url, options);
     options.type = options.type || 'GET';
     options.dataType = options.dataType || 'json';
-    options.context = this;
 
     if (this._shouldSendHeaders(options)) {
       options.headers = this._getFullHeadersHash(options.headers);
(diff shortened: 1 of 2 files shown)


rwjblue:
@jamesarosen - Can you test this fix?

jamesarosen:
Already did: https://github.com/ember-cli/ember-ajax/issues/104#issuecomment-217019134

The `=>` holds on to `this` still, which has a `container`.

rwjblue:
Thank you, I missed that that comment was about this patch.

alexlafroscia:
@jamesarosen I added the fix that you suggested, would you mind testing if that worked?

addon/ajax-request.js
```diff
@@ -42,16 +42,17 @@ function isJSONAPIContentType(header) {
...
 
   constructor() {
     this.init();
   }
 
   init() {
-    this.pendingRequestCount = 0;
     if (testing) {
-      Test.registerWaiter(() => this.pendingRequestCount === 0);
+      Test.registerWaiter(() => (pendingRequestCount === 0));
```
jamesarosen: This [compiles](https://babeljs.io/repl/#?evaluate=true&lineWrap=false&presets=es2015%2Creact%2Cstage-2&experimental=true&loose=false&spec=false&code=Test.registerWaiter(()%20%3D%3E%20(pendingRequestCount%20%3D%3D%3D%200))%3B) to
```js
if (testing) {
  Test.registerWaiter(function () {
    return pendingRequestCount === 0;
  });
}
```
That's totally fine. But when browsers start honoring `=>` natively and we stop compiling, I worry that the callback will hold on to `this` unnecessarily. I don't see anything in the [spec](http://www.ecma-international.org/ecma-262/6.0/#sec-arrow-function-definitions-runtime-semantics-evaluation) that tells me whether this would be true or not. I suspect it _won't_ be a problem since otherwise, every time a programmer did something like `setTimeout(() => { console.log('hi') })` it would leak memory. To be safe, you could write this in ES5 syntax.

jamesarosen:
I ran a few subsets of our test suite and see no lingering `container`s :)

addon/ajax-request.js
alexlafroscia: Yeah, that probably makes more sense. Really good points; might as well do it the "safer" way.

rwjblue:
Right now we are always adding a waiter for each instance, maybe we can just use a single waiter?

Something like the following:

```js
let pendingRequestCount = 0;

if (Ember.Test) {
  Ember.Test.registerWaiter(function() {
    return pendingRequestCount === 0;
  });
}
```

Then throughout the other locations where we use `this.pendingRequestCount++` or `this.pendingRequestCount--` we would just use `pendingRequestCount++` / `pendingRequestCount--`...

alexlafroscia:
Alright, that's what I had before and changed it.  I'll roll back to the previous commit where it's just a global counter.

alexlafroscia:
@jamesarosen @rwjblue alright, I rolled back to the version that 

- Uses a global counter instead of per-instance counters
- Actually seems to fix the problem
- Doesn't mess with `unregisterWaiter`

Does this seem like the best approach?

jamesarosen:
I'm happy as-is. I'm intrigued by @rwjblue's idea of having a single, static waiter. It's not a blocker for our test-suite, though.

alexlafroscia:
Oh yeah, that does make more sense. I mis-read what he was saying earlier.  I'll switch it around to that

rwjblue:
LGTM

PR #107, opened 2016-05-06
Move implementation to a Mixin

### LINKED ISSUE #106: Allow calling _super in init
Right now, if we add a Mixin to all instances of `Ember.Service`, and one of those Mixins has an `init()` function, that `init()` function will never get called by instances of `AjaxService`. 

Example:
```js
let UsefulMixin = Ember.Mixin.extend({
  init() {
    // will never run in AjaxService
    this._pendingThings = [];
  },
  willDestroy() {
    // this will error if init is never run, in particular when AjaxService is torn down
    this._pendingThings.forEach(pendingThing => pendingThing.cancel());
 }
});

Ember.Service.reopen(UsefulMixin);
```

I believe the issue is [here](https://github.com/ember-cli/ember-ajax/blob/master/addon/ajax-request.js#L51-L56) because the code does not call `this._super(...arguments);`. The class will probably have to refactored to be a normal Mixin or Service so that this is possible, rather than an ES6 class.

alexlafroscia:
- Remove the ES6 class
- Move implementation to a Mixin, that can be used in either the service
  or a standalone object
- Closes #106

This shouldn't actually be a breaking change; `ember-ajax/ajax-request` now exports an `Ember` class (rather than ES6 class) that can be used the exact same way that the old ES6 class could be used.


taras:
LGTM 👍

rwjblue:
LGTM

alexlafroscia:
Just as a note: should we consider this a breaking change, where the next release needs to be `3.0`? Or should anyone depending on `AjaxRequest` being an ES6 class still be OK? For what it's worth, none of the tests really had to change to support the Ember class instead, but if someone happened to be using the ES6 class and `extend`, I'm not sure how they would be effected.

rwjblue:
It shouldn't be breaking for those importing the `ember-ajax/ajax-request` and instantiating it with `new` or using `class Foo extends AjaxRequest { }`.

Small demos:

* http://rwjblue.jsbin.com/bobuko/edit?js,output - Using `class Foo extends Ember.Object.extend()`.
* http://rwjblue.jsbin.com/cuhoqo/edit?js,output - Using `new Ember.Object.extend()`.

alexlafroscia:
Oh wow, cool. I didn't know that Ember classes technically work with ES6 `extends`. Thanks for the links @rwjblue!

rwjblue:
@alexlafroscia - Mind making a release for this?

alexlafroscia:
@rwjblue minor or patch bump?

alexlafroscia:
@rwjblue Released as `2.4.0`

PR #112, opened 2016-05-29
Allow configuring namespace on per-request basis

### LINKED ISSUE #111: Namespace for a Single Request
I have a legacy end point that lives at the root namespace of my API (e.g. `/users/signout`) whereas every other request lives in a specific namespace (e.g. `/api/`).

It'd be really nice to be able to temporarily override the namespace for some requests, so that I could do something like:

```javascript
this.get('ajax').del('/users/signout', { namespace: '/' });
```

Where the Ajax service is set to use `/api` globally.

alexlafroscia:
Closes #111

### DIFF
--- a/addon/mixins/ajax-request.js
+++ b/addon/mixins/ajax-request.js
@@ -267,7 +267,7 @@ export default Mixin.create({
     }
 
     const host = options.host || get(this, 'host');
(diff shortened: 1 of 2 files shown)


taras:
👍 LGTM

PR #113, opened 2016-05-31
Add ConflictError class and AbortError handler

alexlafroscia:


alexlafroscia:
Hmm, not sure why those first two commits are in there. They shouldn't be... Both of those commits are on `master` already.

taras:
👍 LGTM

Issue #85, opened 2016-04-08
Integrate with Ember Data out of the box


wagenet:
Right now, unless there's something I've missed, there's no default way to have the ajax service used by Ember Data. As a result, I had to do a work-around like this to avoid duplication:

``` js
// app/adapters/application.js
import Ember from 'ember';
import ActiveModelAdapter from 'active-model-adapter';

const { inject } = Ember;
const { alias } = Ember.computed;

export default ActiveModelAdapter.extend({
  ajaxService: inject.service('ajax'),

  headers: alias('ajaxService.headers')
});
```

It would be awesome for this to happen automatically.

taras:
Totally agree. 

There is an [open issue in Ember Data repo](https://github.com/emberjs/data/issues/4175) and we're hopefully going to start working on this shortly.

alexlafroscia:
I don't think this will ever happen, especially as Ember moves away from `jQuery` as a dependency.

In that world, `ember-ajax` will be an easy way to add `jQuery` only if you want to use `$.ajax`, and even will be a means for importing a custom, slimmed-down build of `jQuery` with only the networking code.

I'm going to close this issue as I don't see it ever actually happening.

taras:
👍

PR #114, opened 2016-06-01
Add mixin for Ember Data integration

alexlafroscia:
@taras and I worked on this PR together, to introduce a mixin that can be added to an Ember Data adapter.  This should be seen as a first step toward solving #85.

When included, the mixin will use the `ajax` service for networking, instead of the code provided as part of the RestAdapter.  Since the `ember-ajax` code was based on the code that it is now replacing, bridging the gap here was pretty straightforward.

The one thing that I'm not sure about right now is the naming of file that the mixin is found in.  I think having it simply called `ajax-support` is a little ambigious, especially when the implementation file is called `ajax-request`.  They're just a little too similar.  I'd love suggestions on a better name.


taras:
This looks great. We need to describe the migration path for people who're looking to adopt this. It can serve as a transition path towards using Ember Ajax with Ember Data. Primarily, we need to show how error handling needs to be adjusted to work with Ember Ajax in Ember Data. 

On that note, we'll need to add Ember Data Error object detection to our error detection helpers.

alexlafroscia:
Yeah, agreed. Definitely need to add some docs, but I'd rather not just add more to the README... There's a *lot* of information in there already.  Maybe usage example in the mixin file?

taras:
What do you think about moving the mixin to `/adapters/ajax-support.js` and creating a README.md file there with this information.

alexlafroscia:
I think that's kind of confusing, having a mixin in the `adapters` directory. It feels like straying from the convention without a strong reason.

taras:
That's kinda how Ember Data does it, Not exactly, but similar https://github.com/emberjs/data/blob/master/addon/-private/adapters/build-url-mixin.js

alexlafroscia:
I see what you're saying... But that's also a "private" module, not public. To me, addons have a given structure and I think adhering to the convention is a better way to go.

taras:
@alexlafroscia ok, what do you think about putting the upgrade instructions into the README for now and later we can give it a section on a documentation site.

taras:
> What's the right way to add detection of the Ember Data errors? We'd need to import the errors and then do an instanceof check, so Ember Data would become an Ember Ajax dependency?

@alexlafroscia that's a great point. I'll think about this today and get back to you.

alexlafroscia:
I'm wondering if it's really necessary to add the ED error checkers... I mean, if they're opting into using the Ember AJAX mixin, they need our error checkers for network errors.  If the ED networking code isn't being run, how do those errors crop up?

taras:
Yeah, you're right. Wi this strategy, it's not necessary because the assumption is that we will be running all of the Ajax code through Ember Ajax with the mixin. We should make this available as beta and have people try it.

alexlafroscia:
Cool.  I'll add a commit with some docs and we can cut a release with this.  Would you want to include this as part of `3.0-beta.1`? Or `2.5`?

taras:
Let's stick within 2.x series for now. So, 2.5-beta.1?

alexlafroscia:
@taras Can I get one last check-over before pulling this?

addon/mixins/ajax-support.js
```diff
@@ -0,0 +1,28 @@
...
+  computed: { alias }
+} = Ember;
+
+export default Mixin.create({
+
+  ajaxService: service('ajax'),
+
+  host: alias('ajaxService.host'),
+
+  namespace: alias('ajaxService.host'),
```
taras: This should alias namespace, right?

alexlafroscia: Good find, thanks!

alexlafroscia:
@taras can I get one last once-over on this before merging?  I fixed that thing you caught.

README.md
```diff
@@ -257,6 +257,32 @@ export default Ember.Route.extend({
...
+import DS from 'ember-data';
+import AjaxServiceSupport from 'ember-ajax/mixins/ajax-support';
+
+const {
+  inject: { service }
+} = Ember;
+const { JSONAPIAdapter } = DS;
+
+export default JSONAPIAdapter.extend(AjaxServiceSupport, {
+  ajaxService: service('ajax'),
```
taras: The mixin does this, we don't need this line and we need to move `namespace: `api`` to  the service. Right?

taras:
Other than the 1 comment, it's good to go.

README.md
alexlafroscia: RE: the `ajaxService` line, I know we don't need it.  I just wanted to illustrate that you can specify the service to use, but maybe that was confusing? What do you think @taras?

RE: the `namespace` line, I wanted to show how you can specify certain things on the adapter.  Otherwise, the whole adapter is just

```javascript
export default JSONAPIAdapter.extend(AjaxServiceSupport);
```

which isn't much to show, really...

taras: Yeah, I think that's fine. We want to show people that they can allow `ember-ajax` to take care of everything.

alexlafroscia: Alright, I'll update the example code to show a "minimal" example, and also include an "extended" one that's similar to what I originally put.  Sounds good?

dwickern:
This appears to add the namespace twice: first in the ember-data adapter and again in the ember-ajax service.

Issue #79, opened 2016-03-31
Send "Content-Type": "application/json; charset=utf-8" by default


nucleartide:
Hey folks, first of all great job with the library :)

So I notice that ember-ajax sends data with a content type of `application/x-www-form-urlencoded` by default. Since sending JSON is so common, would you think it's a good idea for `application/json` to be the default content type instead? (Like in [superagent](https://visionmedia.github.io/superagent/#post-/-put-requests).)

For example, instead of this:

```js
this.get('ajax').post(url, {
  contentType: 'application/json; charset=utf-8',
  data: JSON.stringify({
    hello: 'world',
    foo: 'bar'
  })
}
```

You could do something like this:

```js
this.get('ajax').post(url, {
  data: {
    hello: 'world',
    foo: 'bar'
  }
})
```

I'm thinking it would just cut down on some boilerplate. What do you think?

alexlafroscia:
I'm absolutely in favor of doing this.  It might break some people's default behavior though, if they're relying on the fact that the default `contentType` is the same as what `jQuery` uses.  I agree completely that `application/json` is a more sensible default, though.

If you'd like to submit a PR for this in the next day or so (with a test) then I'd like to pull that in before the 2.0 release, which has some other breaking changes.  Because of the breaking nature of the change, though, if it doesn't make it into 2.0 (which we're planning to make official _very_ soon) then it would have to wait until 3.0.

timkendall:
Just ran into this myself. Would be nice if we had a way to just set this when we extend the `ajax` service.

cqcsobreyra:
Just wanted to share an idea for a quick dirty fix how to achieve default options for those who might find it useful. In the ajax service you could do something like

```
import Ember from 'ember';
import AjaxService from 'ember-ajax/services/ajax';

export default AjaxService.extend({
    configuredRequest(url, options){
        var defaultOptions = {
            cache: false,
            contentType: 'application/json; charset=UTF-8',
            processData: false
        }

        if (options) {
            $.extend(defaultOptions, options);
        }
        this.request(url, options);
    }
});

```
And then just call this.get('ajax').configuredRequest('/yoururl', yourOptions);

zeiv:
[...]

PR #116, opened 2016-06-05
Add service property for default contentType

alexlafroscia:
This PR is closesly related to #79.  Since going and changing the default content type would be a breaking change, we're planning to make that happen for version `3.0`.  For the time being, this PR allows the service to set a `contentType` property that will be used by default, which itself defaults to the same value that jQuery itself would use.  This makes the default value much more explcit, which will hopefully help mitigate errors due to the default not being clear, as well as provides a clear pathway for changing the default value in the next major version.

### DIFF
--- a/addon/mixins/ajax-request.js
+++ b/addon/mixins/ajax-request.js
@@ -76,6 +76,17 @@ if (testing) {
 
 export default Mixin.create({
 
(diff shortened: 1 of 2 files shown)


taras:
LGTM +1

PR #119, opened 2016-06-10
Case-insensitive handling of headers

sandstrom:
https://www.w3.org/Protocols/rfc2616/rfc2616.html

### DIFF
--- a/addon/mixins/ajax-request.js
+++ b/addon/mixins/ajax-request.js
@@ -44,7 +44,7 @@ const JSONAPIContentType = 'application/vnd.api+json';
 function retrieveHeader(headers, name) {
   let found = null;
   
-  Object.keys(headers).find((key) => {
+  Object.keys(headers).forEach((key) => {
     let value = headers[key];
     
     if (value.toLowerCase() === name.toLowerCase()) {
--- a/addon/mixins/ajax-request.js
+++ b/addon/mixins/ajax-request.js
@@ -42,16 +42,13 @@ const JSONAPIContentType = 'application/vnd.api+json';
 // Header names are case-insensitive
 // https://www.w3.org/Protocols/rfc2616/rfc2616.html
 function retrieveHeader(headers, name) {
-  let found = null;
-  
+  var found = null;
   Object.keys(headers).forEach((key) => {
     let value = headers[key];
-    
     if (value.toLowerCase() === name.toLowerCase()) {
       found = value;
     }
   });
-  
   return found;
 }
 
--- a/addon/mixins/ajax-request.js
+++ b/addon/mixins/ajax-request.js
@@ -39,11 +39,27 @@ const {
 } = Ember;
 const JSONAPIContentType = 'application/vnd.api+json';
 
+// Header names are case-insensitive
+// https://www.w3.org/Protocols/rfc2616/rfc2616.html
+function retrieveHeader(headers, name) {
+  let found = null;
+  
+  Object.keys(headers).find((key) => {
+    let value = headers[key];
+    
+    if (value.toLowerCase() === name.toLowerCase()) {
+      found = value;
+    }
+  });
+  
+  return found;
+}
+
 function isJSONAPIContentType(header) {
   if (isNone(header)) {
     return false;
   }
-  return header.indexOf(JSONAPIContentType) === 0;
+  return header.toLowerCase().indexOf(JSONAPIContentType) === 0;
 }
 
 function startsWithSlash(string) {
@@ -107,7 +123,7 @@ export default Mixin.create({
       url: hash.url
     };
 
-    if (isJSONAPIContentType(hash.headers['Content-Type']) && requestData.type !== 'GET') {
+    if (isJSONAPIContentType(retrieveHeader(hash.headers, 'Content-Type')) && requestData.type !== 'GET') {
       if (typeof hash.data === 'object') {
         hash.data = JSON.stringify(hash.data);
       }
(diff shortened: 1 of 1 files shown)


alexlafroscia:
This change looks good to me, but the tests aren't happy about it.  If you could take a look at get them passing, I'd be happy to merge this in.

sandstrom:
Ah! I was about to use `find` first (ES6/7), but then reverted to a basic `forEach` loop — but forgot to change the actual method… :)

alexlafroscia:
@sandstrom Looks like there are a few errors still, one JSCS issue and a few tests that now fail.

sandstrom:
@alexlafroscia JSCS "errors" are solved. Any thoughts on this other issue?

addon/mixins/ajax-request.js
```diff
@@ -42,16 +42,13 @@ const JSONAPIContentType = 'application/vnd.api+json';
 // Header names are case-insensitive
 // https://www.w3.org/Protocols/rfc2616/rfc2616.html
 function retrieveHeader(headers, name) {
-  let found = null;
-  
+  var found = null;
```
alexlafroscia: This should still use `let`, not `var` for a mutable variable.

addon/mixins/ajax-request.js
```diff
@@ -39,11 +39,24 @@ const {
 } = Ember;
 const JSONAPIContentType = 'application/vnd.api+json';
 
+// Header names are case-insensitive
+// https://www.w3.org/Protocols/rfc2616/rfc2616.html
+function retrieveHeader(headers, name) {
+  var found = null;
+  Object.keys(headers).forEach((key) => {
```
alexlafroscia: This can be simplified using `.find`

```js
return Object.keys(headers).find((key) => {
  const value = headers[key];
  return value.toLowerCase() === name.toLowerCase();
});
```

The `.find` is sort of like `.filter`, which returns an array of only the elements in the original array for which the callback returns `true`.  However, it's different in that it only returns the first element in the array that matches, which is what it seems like you're going for here.

addon/mixins/ajax-request.js
sandstrom: You are right. I thought `var` was needed to set the value from within a child-scope (since block vs. function scoping is the difference between the two), but reading up on the docs I think `let` will work, i.e. it's still available within child-blocks, only the shadowing is different between the two.

addon/mixins/ajax-request.js
sandstrom: I tried find, but that isn't supported in PhantomJS (and IE11).

alexlafroscia: Hmm, I see. Maybe we could use the one included in the [`Ember.A()` class](http://emberjs.com/api/classes/Ember.Array.html#method_find)?

It looks like some of the issues on Travis are style-related (we shouldn't be using `var`).  I'm not sure about the other ones, I have the branch pulled locally but need to dig into it to figure out why it's failing now.  I'll be doing that when I get home from work tonight.

addon/mixins/ajax-request.js
```diff
@@ -39,11 +39,24 @@ const {
...
+    let value = headers[key];
+    if (value.toLowerCase() === name.toLowerCase()) {
```
alexlafroscia: At least one of the failing tests is due to this... We don't actually want to compare the value behind the key to the desired key, we want to match against the key.  I'm going to try to push a commit to this PR to get things passing.

alexlafroscia: @sandstrom Actually, before I go ahead and do this myself, do you want to take a stab at fixing this?  We need to compare the _key_ to the _name_ parameters, and then return the corresponding value, not compare against the value.

alexlafroscia:
Instead of implementing the `retrieveHeader` file as a function within the current file, I'd like to move it into `addon/util/retrieve-header.js` and then have a test file that ensures that it works correctly.  That'll make bugs like the ones we've seen here much easier to solve in the future.

The things I'm seeing as issues, just to put them all together, are:

- [ ] Refactor the `retrieveHeader` function into a separate utility file, with tests
- [ ] Compare the provided key against a key in the headers object, not the value.  Then, return the corresponding value
- [ ] Switch back to using `.find`, even if that means using one of the Ember array helper classes.  It makes the code a bit cleaner as `.find` expresses the intended behavior exactly.
- [ ] Remove any references to `var` and replace with `let` (although this should just implicitly happen when the above bullet point is fixed

I have a branch locally that has some of that implemented already, but I want to give you a chance to fix it up first.  If you'd rather I clean things up, I'm happy to do that too!

addon/mixins/ajax-request.js
alexlafroscia: Gotcha.  All of that is a little too stateful for my taste.  The approach I took is as follows, but feel free to do it however you like:

```js
  const matchingKey = Ember.A(Object.keys(headers)).find((key) => {
    return key.toLowerCase() === desiredKey.toLowerCase();
  });
  if (matchingKey) {
    return headers[matchingKey];
  }
```

alexlafroscia:
Hey @sandstrom, any movement on your PR here?

sandstrom:
@alexlafroscia Thanks for pinging! I'd forgotten about this.

Feel free to run with this, I don't have any bandwidth right now.

alexlafroscia:
@sandstrom no problem! Thanks for the update.

alexlafroscia:
@taras what do you think about closing this PR but adding the intended fix (always matching against headers case-insensitively) to the `v3.0.0` tasks?  We could pretty simply match using a regex and just make that case insensitive.

taras:
Sounds like a good plan

PR #121, opened 2016-06-16
Move to tests to Mocha

alexlafroscia:
Moves the tests over to Mocha instead of QUnit.  I think that the tests will be a little easier to organize using Mocha instead through the nesting that it provides.

Also, update Suave to the latest version, which is compatible with Mocha.

- [ ] Refactor tests to leverage nesting `describe` blocks for clarity
- [ ] Try to figure out of the error-catching test can be enabled (probably need someone with experience with Mocha)


tests/.eslintrc.js
```diff
@@ -1,5 +1,8 @@
 module.exports = {
   env: {
     'embertest': true
+  },
+  globals: {
+    'expect': true
```
Turbo87: you can probably just use the `mocha` env here which also includes `describe`, `it`, ...

Turbo87: also `expect` itself should probably be imported from `chai` instead of using the global, right?

alexlafroscia: Good notes all around; I don't think I actually use the `expect` global anywhere so removing it seems like the right way to go.

alexlafroscia:
I'm going to merge this, now that it won't conflict with `master` and all the tests are passing.  There are still some things to do, but I think they're both minor and I want to prevent anyone else from branching from the project while it still uses QUnit since that makes merging with this branch really problematic..

PR #126, opened 2016-06-27
Allow payload response to be a falsey but non-null or undefined value

SteelBurgher:


alexlafroscia:
Could you add a unit test instead of a component/integration test?  I don't think there's a reason we need to invoke all of the component/DOM logic to verify that this works correctly... Unless I'm missing something.

I'd like to see a few test cases added, too: `undefined`, `null`, and some falsey value (like `0` that you picked) to make sure they all work like we expact.

SteelBurgher:
I totally agree. Good unit tests forthcoming.

alexlafroscia:
Fantastic 👍

tests/unit/ajax-request-test.js
```diff
@@ -647,6 +647,25 @@ test('it JSON encodes JSON:API "extension" request data automatically', function
...
+  const service = new AjaxRequest();
+
+  const payloadWithFalseyString = service.handleResponse(200, {}, '');
+  assert.equal(payloadWithFalseyString, '');
+
+  const payloadWithFalseyNumber = service.handleResponse(200, {}, 0);
+  assert.equal(payloadWithFalseyNumber, 0);
+
+  const payloadWithNaN = service.handleResponse(200, {}, NaN);
+  assert.equal(isNaN(payloadWithNaN), true);
```
alexlafroscia: You can just do this as `assert.ok`.  Not a merge-blocker, just a note.

SteelBurgher:
I switched to using assert. what's with the appveyor build always failing?

alexlafroscia:
I'm not sure.  It's not just your build, all of the PRs are failing like this.  Not sure if it's an Appveyor issue or a Windows issue... I'm going to get set up in a Windows VM tonight and see what's going on, because I'm not comfortable pulling this if there is actually an issue on Windows.  Hold tight and we'll get this pulled in as soon as I know what's up there.

alexlafroscia:
Actually, I just noticed that my PR from last night was fine on Appveyor.. Can you try rebasing off of the current master and pushing everything again?  Maybe it was a problem with a dependency that is now updated.

alexlafroscia:
It seems like the tests are passing again, on the current `master` branch.  However, that branch moves the test runner over to Mocha instead of Qunit, so you'll have to update your tests.  To save you the trouble, if you add me as a collaborator on your fork, I'll handle the rebase/test update and push a commit up to this PR so we can get this closed.

SteelBurgher:
I added you as a collaborator. Sorry was away for some time on a long vacation.

alexlafroscia:
No worries 😄  I'll try to get this finished up today.

alexlafroscia:
@SteelBurgher Can you check out my version of your branch over [here](https://github.com/alexlafroscia/ember-ajax/tree/payload-checking-modification)?  It should be exactly the same, but with the (same) tests written in Mocha and rebased off the current `master`.  The tests fail, but that's problem on `master` right now too so don't worry about that.  I just wanted to verify that it looks good to you, I didn't want to go ahead and `push --force` your branch.

SteelBurgher:
@alexlafroscia looks good, you can push.

alexlafroscia:
Test are failing but it's due to [this Ember bug](https://github.com/emberjs/ember.js/issues/13888) so I'm not worried about it.

PR #128, opened 2016-07-01
Refactor request logic to avoid duplicate calls/Refactor tests

### LINKED ISSUE #122: Options method getting called twice
The method `raw` in het `ajax-request` mixin is calling `this.options`, while `request` is already calling this method, seems unneccessary to me? The reason I'm asking is because I'm overriding the `options` method in my custom addon, which will result is some wrong behaviour because its being called twice (second time with an already modified options object from the first time).

Can't `raw` simply get the `hash` calculated within the `request` method?

alexlafroscia:
We used to call `options()` twice, but avoiding that extra work isn't too hard.  Also adds tests to verify that the method is only called once per type of request method called (`request` and `raw`, respectively).

Closes #122

***

Also adds a bunch of test refactoring to make them a little bit cleaner and to take advantage of some of the Mocha and Chai features that we're not yet using.


alexlafroscia:
@rwjblue would you mind taking a look real quick, before I merge this?

PR #129, opened 2016-07-04
if user specifies json-api contentType, jsonify data object

koriroys:
This is a follow-up to #116, which added the overridable contentType property. The current issue is, even if I set the contentType to "application/vnd.api+json", the data body remains non-jsonified. This is my  attempt to address that issue. Is this the right way to go about this? Feedback welcome. Also, I'm guessing you'll want some tests?

### DIFF
--- a/addon/mixins/ajax-request.js
+++ b/addon/mixins/ajax-request.js
@@ -34,8 +34,9 @@ const {
   RSVP: { Promise },
   Test,
   get,
+  isEqual,
   isArray,
-  isNone,
+  isPresent,
   merge,
   run,
   runInDebug,
@@ -44,11 +45,17 @@ const {
 } = Ember;
 const JSONAPIContentType = 'application/vnd.api+json';
 
-function isJSONAPIContentType(header) {
-  if (isNone(header)) {
-    return false;
+function isJSONAPIContentType(hash) {
+  let header = hash.headers['Content-Type'];
+  let defaultContentType = hash.contentType;
+
+  if (isPresent(header) && isEqual(header.indexOf(JSONAPIContentType), 0)) {
+    return true;
   }
-  return header.indexOf(JSONAPIContentType) === 0;
+  if (isEqual(defaultContentType, JSONAPIContentType)) {
+    return true;
+  }
+  return false;
 }
 
 function startsWithSlash(string) {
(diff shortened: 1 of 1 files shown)


addon/mixins/ajax-request.js
```diff
@@ -222,7 +229,7 @@ export default Mixin.create({
       url: hash.url
     };
 
-    if (isJSONAPIContentType(hash.headers['Content-Type']) && requestData.type !== 'GET') {
+    if (isJSONAPIContentType(hash) && requestData.type !== 'GET') {
```
alexlafroscia: I personally think it would make more sense to leave the `isJSONAPIContentType` function alone and instead just use it twice, something like

```javascript
if ((isJSONAPIContentType(hash.contentType) || isJSONAPIContentType(hash.headers['Content-Type']) && request.type !== 'GET') {
  ...
}
```

if we need to check both properties.

alexlafroscia:
Totally makes sense; this is definitely a change we want to make.

I left a comment in the code; in addition, we definitely want to add some tests to verify that the content is JSON-ified with the JSON:API content type specified as either the `contentType` property _or_ in the headers.

koriroys:
Actually ran into some issues trying to implement your requested changes. Problems pop up if you just `or` the two fields. E.g., what should happen if the user specifies `contentType` as form encoded, but specifies JSON:API in the Content-Type header? Should we warn the dev in the case they don't match? Or would it make more sense to just set the Content-Type header to JSON:API if the dev sets `contentType` to jsonapi?

alexlafroscia:
> Should we warn the dev in the case they don't match

Hmm... That's interesting. I hadn't thought about that.

I like the idea of issuing a warning or error; a mismatch there probably means that the developer has some conflicting configuration that they may not realize.

yads:
Just curious, why only do this on jsonapi content types? Why not also do it for application/json content types? I just recently ran into an issue where I integrated AjaxServiceSupport into my application adapter (which extends RESTAdapter). I had to override the service request method to have the same functionality as https://github.com/emberjs/data/blob/99ffd78566af19d98f8cad2231731e0e2c9ec603/addon/adapters/rest.js#L1071

alexlafroscia:
@koriroys Hey Kori, do you have any updates on this?

koriroys:
I just let it fall off my radar. Actually don't have access to the code where I ran into this issue. What changes do you want to get this in?

alexlafroscia:
@koriroys I left this comment a while ago https://github.com/ember-cli/ember-ajax/pull/129#r69491493

alexlafroscia:
Due to inactivity and the fact that #240 also deals with this code, I'm going to close this. If that's wrong, I'm happy to re-open later.

PR #135, opened 2016-08-02
Document stand-alone usage

### LINKED ISSUE #134: Access in utilities
I can't find a way to access ember-ajax inside a utility. Is this possible, if not, why? This requires me to either wrap `Ember.$.ajax` or include `ic-ajax` or some other library. Seems wasteful.

knownasilya:
Fixes #134

### DIFF
--- a/README.md
+++ b/README.md
@@ -280,6 +280,15 @@ That's all the configuration required!  If you want to customize the adapter, su
 
 Note that instead of using the Ember Data error checking code in your application, you should use the ones provided by Ember AJAX.
 
+## Stand-Alone Usage
+
+If you aren't using Ember Data and do not have access to services, you
+can import the ajax utility like so:
+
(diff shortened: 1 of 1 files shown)


alexlafroscia:
If we're going to mention this in the readme (and I'm worried out readme is getting really bloated) we should do both `request` and `raw`, and explain the difference.

knownasilya:
@alexlafroscia @rwjblue how is this?

alexlafroscia:
Great! Can you just add a little usage example below each import?  Something like

```js
import request from 'ember-ajax/request';

return request('/foo', { ... });
```

and the same for raw.  Just in case someone doesn't know what to do once they have the function.

knownasilya:
Added examples

alexlafroscia:
Thank you!

PR #137, opened 2016-08-05
Add whitelist of files to publish

alexlafroscia:


stefanpenner:
@taras mind confirming, and if its good releasing?

taras:
@stefanpenner yes, look good. This is an initiative that happening across all addons?

alexlafroscia:
@taras Yeah, @stefanpenner made a list of addons that get installed automatically and how many files there are in them.  `ember-ajax` was on the list.  We didn't have too many files that could be left out but I figured we should still do our part and lock things down as much as possible.

PR #139, opened 2016-08-05
add failed test for duplicate namespace

dwickern:
If I set the namespace in the ajax service, and I use ember-data integration from #114, the mixin copies the namespace into the application adapter. The namespace is added to the URL in both places and gets repeated:
```
not ok 2 PhantomJS 2.1 - Acceptance | ember data integration can set the namespace for all ajax requests
    ---
        message: >
            Pretender intercepted GET /api/api/posts/1 but no handler was defined for this type of request
```
Workaround: explicitly set `namespace: null` in the application adapter

### DIFF
--- a/tests/acceptance/ember-data-integration-test.js
+++ b/tests/acceptance/ember-data-integration-test.js
@@ -6,6 +6,7 @@ const { equal } = assert;
 import destroyApp from 'dummy/tests/helpers/destroy-app';
 import startApp from 'dummy/tests/helpers/start-app';
 
+import AjaxService from 'ember-ajax/services/ajax';
 import Pretender from 'pretender';
 import { jsonResponse } from 'dummy/tests/helpers/json';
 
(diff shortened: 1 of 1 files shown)


alexlafroscia:
Awesome, thank you for making a failing test case! That makes my job so much easier.  I'll take a look at this this weekend.

alexlafroscia:
Hmm, this is an interesting problem.  Turns out that Ember Data builds the URL before calling our `ajax` method, and I don't think there's anything we can do to avoid that... So we need to update our `._buildURL` to avoid adding the namespace and host if called from ED since it will already be in the URL.

alexlafroscia:
Sorry about the long delay -- I finally got to putting a fix together!  Once that PR is accepted we can close this PR; the new PR includes your commit here to add the failing test.

dwickern:
Thanks!

alexlafroscia:
Closing in favor of #153

PR #146, opened 2016-08-19
fix multiple calls to the options method when using the ajax-support mixin

kketch:
The ajax-support mixin currently calls ember-data adapter's `ajaxOptions` method which is overriden by the same mixin and calls the `options` method. But the `options` method is then again called by the ajaxService, as a result it is called twice.

To fix this ajaxOptions is not overriden anymore in the ajax-support mixin, and is
not called anymore by the ajax method, as it should not be needed by ember-ajax I think
and is a private method of the ember-data adapter.

### DIFF
--- a/addon/mixins/ajax-support.js
+++ b/addon/mixins/ajax-support.js
@@ -34,13 +34,8 @@ export default Mixin.create({
    */
   headers: alias('ajaxService.headers'),
 
-  ajax(url, type, options) {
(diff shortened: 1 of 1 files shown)


alexlafroscia:
Hey, thanks for submitting this.  I think this makes a lot of sense, since `ajaxOptions` doesn't actually get called anywhere other than the `ajax` method we're overwriting anyway!

PR #153, opened 2016-09-08
Fix adding namespace twice

alexlafroscia:
If the URL given already starts with the namespace, don't both to build the URL again.  This prevents issues with Ember Data, where the URL given to the request method is pre-built.

Implements a fix for the issue reported in #139


taras:
LGTM 👍

jherdman:
Would it be possible to cut a release with this fix?

alexlafroscia:
Oops, yeah. Didn't realize that that never happened.

alexlafroscia:
@jherdman released `v2.5.2` just now

Issue #159, opened 2016-09-12
$.parseJSON deprecation


jcbvm:
With jquery 3+, `$.parseJSON` is deprecated, they suggest to just use `JSON.parse`. Is there any reason why the jquery method is used instead of just `JSON.parse`?

taras:
@jcbvm it's probably an oversight. Can you create a PR with this change?

jcbvm:
@taras see https://github.com/ember-cli/ember-ajax/pull/160

alexlafroscia:
Closed by #160

PR #160, opened 2016-09-12
Replaced $.parseJSON with JSON.parse. #159

jcbvm:
The method `$.parseJSON` has been deprecated in jQuery 3+, they suggest to replace it with `JSON.parse`. This pull request does the suggested replace. This should not break any code because `JSON.parse` is supported by all major browsers and IE8+.

### DIFF
--- a/addon/mixins/ajax-request.js
+++ b/addon/mixins/ajax-request.js
@@ -776,7 +776,7 @@ export default Mixin.create({
    */
   parseErrorResponse(responseText) {
     try {
-      return $.parseJSON(responseText);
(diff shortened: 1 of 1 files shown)


taras:
It seems that `$.parseJSON` uses `JSON.parse` when it's available http://stackoverflow.com/a/10362313

@alexlafroscia is this a breaking change?

alexlafroscia:
Assuming that the test suite still passes, I image it's probably fine to change this. It seems like all of the browsers we care about support `JSON.parse`.

Issue #167, opened 2016-09-29
Ability to access the json response in case of error


gnapse:
The documentation mentions that error handling is done in the `catch` handler of the promise. This handler receives an instance of an error object. But there's no way to access the json that came in the error response body. At least there's no documented way that I could find, and I tried checking if the `catch` handler received a second argument, or if the error object contains this somehow, but I couldn't find it.

alexlafroscia:
Yeah... it's admittedly not a great system right now. If you need access to the original response you can use `raw`.

gnapse:
I am willing to work on a pull request for introducing this capability, but I would need to know first what would be a good way to expose the error response in all situations. Are there already some ideas? Or perhaps we can discuss here?

Like I said, it could be a second argument to the `catch` method (not sure how compliant is this with the promises standard):

```javascript
// additional parameter
this.get('ajax').request(url)
  .then(response => this.handleResponse(response))
  .catch((error, response) => console.log('error response', response);
```

Alternatively, response access can be given via the error object:

```javascript
// via the error object
this.get('ajax').request(url)
  .then(response => this.handleResponse(response))
  .catch(error => console.log('error response', error.response));
```

alexlafroscia:
I think right now, the intention was that if you needed "all the information" then to use `raw`.  However, I'm open to discussing another approach.  Maybe we can pull in @taras and see how he feels about chaning this up.

gnapse:
I read about the `.raw` method in the README, did not know about it before. Is this accessible as a method of the service object? Because I have no objection to use it like this:

```javascript
this.get('ajax').raw(url, options)
  .then(...)
  .catch(...);
```

Does it still preserve any headers or customizations I made on the ajax service (namespace, headers, etc.)? If that's so, then this case is mute. The thing is, in the README it does not appear this is the case. They mention importing it as some lower level API that does not use the service object at all, but rather the service object uses it.

alexlafroscia:
Sure is!

[This](https://github.com/ember-cli/ember-ajax/blob/b32b9ee5ede69e09a162eb98d354f7de6297f679/addon/mixins/ajax-request.js#L201-L206) is the only difference between `request` and `raw`, all the internals are the same.  Sorry that the docs aren't clear that it's a method as well.

If you're good with using `raw` then I'm going to close this.  LMK if that doesn't fix things for you.

PR #168, opened 2016-10-04
Document how to access the full response on error

gnapse:
This addresses the concerns exposed in #167 about being able to access the full json response in case of a failed request.

### DIFF
--- a/README.md
+++ b/README.md
@@ -263,6 +263,18 @@ export default Ember.Route.extend({
 
 If your errors aren't standard, the helper function for that error type can be used as the base to build your custom detection function.
 
+#### Access the response in case of error
+
+If you need to access the json response of a request that failed, you can use the `raw` method instead of `request`.
+
+```js
+this.get('ajax').raw(url, options)
+  .then(response => this.handleSuccess(response))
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -263,6 +263,18 @@ export default Ember.Route.extend({
 
 If your errors aren't standard, the helper function for that error type can be used as the base to build your custom detection function.
 
+#### Access the response in case of error
+
+If you need to access the json response of a request that failed, you can use the `raw` method instead of `request`.
+
+```js
+this.get('ajax').raw(url, options)
+  .then(response => this.handleSuccess(response))
```
alexlafroscia: I'm fairly sure that the first argument here should be `{ response }` right?

gnapse: I thought the `.catch` part was the only one receiving all the extra info aside from the response. But if `.then` also receives the same as `.catch`, then yes, you're right. I will confirm from the codebase, but if you can positively confirm before, let me know.

alexlafroscia: Both `.catch` and `.then` work the same way in this instance.

gnapse: Ok, fixed. Thanks for the info.

PR #173, opened 2016-10-11
ConflictErrors should return the errors payload from the server.

joshuabremer:


addon/errors.js
```diff
@@ -102,8 +102,8 @@ AbortError.prototype = Object.create(AjaxError.prototype);
  * @public
  * @extends AjaxError
  */
-export function ConflictError() {
-  AjaxError.call(this, null, 'The ajax operation failed due to a conflict');
+export function ConflictError(errors) {
+  AjaxError.call(this, errors, 'The ajax operation failed due to a conflict');
```
joshuabremer: I think this was an oversight. All the other 4** errors do this. However this is placed below Abort and Timeout error which probably made someone thing this is a purely client-side generated error.

tests/unit/mixins/ajax-request-test.js
```diff
@@ -906,6 +906,8 @@ describe('Unit | Mixin | ajax request', function() {
           .catch(function(reason) {
             expect(reason instanceof errorClass).to.be.ok;
             expect(reason.errors && typeOf(reason.errors) === 'array').to.be.ok;
+            expect(reason.errors[0].id).to.equal(1);
+            expect(reason.errors[0].message).to.equal('error description');
```
joshuabremer: If you'd like this logic elsewhere, that's fine. But I figured since you already had an assertion that the `reason.errors` is an array, i thought i'd add to that asserting that the content is correct.

alexlafroscia:
Yup, definitely an oversight.  Thanks for the fix!

PR #176, opened 2016-10-17
Refactor Error Classes

### LINKED ISSUE #175: normalizeErrorResponse should do less
Our original intention with `normalizeErrorResponse` was to provide a standard way of presenting error responses. To achieve this goal we detected different scenarios and coerced the errors into JSONAPI format. `normalizeErrorResponse` is now quite unpredictable, especially to new users.

To goal is to reduce the amount of work that `normalizeErrorResponse` does to allow users to opt-in to formatting their error response instead of doing it manually.
# TODO
- [x] Add a deprecation warning to `normalizeErrorResponse` stating that current behavior of `normalizeErrorResponse` is deprecated and will be removed in Ember Ajax 3.0
- [x] Refactor Error Classes to ES6 and add errors getter that throws a deprecation warning that `errors` property will be removed in Ember Ajax 3.0
- [x] Copy the payload onto `payload` property of error instance
# Related Issues
- https://github.com/ember-cli/ember-ajax/issues/157

jherdman:
This breaking change accomplishes a few things:

* Deprecates the AjaxError#errors property. Going forward we'll do much
  less with errors, and instead simply provide the payload from the
  server's error response. The payload will be accessible via the
  #payload property.
* Refactors Ajax Error, and friends, into ES6 classes. This is
  potentially a breaking change as we no longer inherit from
  Ember.Error.

Closes #175

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,7 @@ env:
   - EMBER_TRY_SCENARIO=ember-release
   - EMBER_TRY_SCENARIO=ember-beta
   - EMBER_TRY_SCENARIO=ember-canary
+  - EMBER_TRY_SCENARIO=ember-1-13
 
 matrix:
   fast_finish: true
--- a/addon/errors.js
+++ b/addon/errors.js
@@ -4,17 +4,13 @@ const { Error: EmberError } = Ember;
 
 /**
  * @class AjaxError
- * @private
+ * @public
+ * @extends Ember.Error
  */
-export function AjaxError(errors, message = 'Ajax operation failed') {
+export function AjaxError(payload, message = 'Ajax operation failed') {
   EmberError.call(this, message);
 
-  this.errors = errors || [
-    {
-      title: 'Ajax Error',
-      detail: message
-    }
-  ];
+  this.payload = payload;
 }
 
 AjaxError.prototype = Object.create(EmberError.prototype);
@@ -24,8 +20,8 @@ AjaxError.prototype = Object.create(EmberError.prototype);
  * @public
  * @extends AjaxError
  */
-export function InvalidError(errors) {
-  AjaxError.call(this, errors, 'Request was rejected because it was invalid');
+export function InvalidError(payload) {
+  AjaxError.call(this, payload, 'Request was rejected because it was invalid');
 }
 
 InvalidError.prototype = Object.create(AjaxError.prototype);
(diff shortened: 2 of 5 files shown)


alexlafroscia:
Thanks @jherdman! This all looks 👌  to me.

I think we should make a `3.0.0` branch to merge these things to for now, so that we can land non-breaking changes for the time being (like the deprecation warnings when using the existing error normalization method) without having to ship these breaking ones, too.  Does that sound good?  If so, update this PR to merge into `3.0.0` (which I'll make right now).

jherdman:
Shouldn't this merge before 3.0? I believe the release target should be 2.6. If not, I may need to update the target version this deprecation will be removed.

alexlafroscia:
If it's a breaking change then it can't land before `v3.0.0` right?

jherdman:
My apologies, you're correct. Before we merge this we ought to decide when we'll remove the `errors` property to update the [`until:`](https://github.com/jherdman/ember-ajax/blob/bbb5865132f4ed3a2d79cb73ebfb55797acd00d1/addon/errors.js#L36) tag.

jherdman:
@taras our snazzy fixup has been pushed!

jherdman:
Fails look like rubbish. @taras are you able to kick the builds?

taras:
@jherdman I don't think I can. Can you please push an empty commit with: `git commit -m "Kicking CI" ---allow-empty`

jherdman:
Gah! AppVeyor is being a jerk. I'll investigate the fails today.

alexlafroscia:
I've been thinking lately that we should just drop AppVeyor. It's constantly slow and unstable, and we don't do anything that should be different with Broccoli that should be different between Windows and Linux. Every time I've seen a failure on there, it's just AppVeyor being a butt and not an actual Windows issue.

jherdman:
@alexlafroscia yeah, totes. Especially if we're just building against Chrome on Windows.

jherdman:
OK, so Ember 1.13 is failing consistently on AppVeyor. One fail, fine, two? Not so cool. Three: we've got a problem.

I've managed to pooch something on my machine tonight, but I'll do my best to look into this soon.

taras:
@jherdman yeah, that's weird. Gotta checkout in 1.13 and figure out why it's dying.

You can do: `ember try ember-1.13 --skip-cleanup` and it'll switch to 1.13

jherdman:
@alexlafroscia do you mind updating the 3.0 branch? There are some spurious commits in this.

@taras so, the most recent commit fixes the fails for 1.13. It seems like 1.13 always raises on deprecation. I thought about detecting 1.13 and telling it not to throw, but that seemed more complicated than actually fixing some tests now. Thoughts?

jherdman:
CI is ruining me...

jherdman:
Ping @alexlafroscia @taras . If y'all are happy, I'll squash this down and push it back up.

taras:
No worries, merged :)

jherdman:
💃

taras:
Thank you for helping us push this through :)

PR #186, opened 2016-10-28
Use `\n` as CLRF in FastBoot mode to parse headers

### LINKED ISSUE #162: Parse headers not working in FastBoot
Hi,

I found a problem with parsing headers in FastBoot. Headers are [split](https://github.com/ember-cli/ember-ajax/blob/master/addon/utils/parse-response-headers.js#L1) with `\u000d\u000a`, but in my case this results in `[ 'server: nginx/1.4.6 (Ubuntu)\ndate: Wed, 14 Sep 2016 11:46:55 GMT\ncontent-type: application/json;charset=UTF-8\nt...]`.

I had to replace `\u000d\u000a` with `\n` to make this working, can make PR if it's OK.

bobisjan:
fix #162

### DIFF
--- a/addon/utils/parse-response-headers.js
+++ b/addon/utils/parse-response-headers.js
@@ -1,4 +1,6 @@
-const CLRF = '\u000d\u000a';
+import isFastBoot from './is-fastboot';
+
+const CLRF = isFastBoot ? '\n' : '\u000d\u000a';
 
 export default function parseResponseHeaders(headersString) {
   let headers = {};


bobisjan:
I'm not sure about test, because https://github.com/kaliber5/ember-fastboot-addon-tests requires Node.js 4+.

alexlafroscia:
I believe that @rajzshkr was working on a PR that would fix this, as well as add tests.  Can you weigh in, Raja?

alexlafroscia:
So, I've been thinking about this some and I feel like `\n` being required in Fastboot is probably a bug in `najax`.

For one, `CLRF` is an actual constant, so I feel like we shouldn't be defining it as `\n`.  Additionally, the JS docs say that the headers should be broken up by `CLRF` [here](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/getAllResponseHeaders), so if Node behavior is different than that, it's probably a bug somewhere upstream

alexlafroscia:
Based on my previous comment and lack of movement here, I'm going to close this. We can re-open if it becomes clear that we should actually be using `\n` instead.

PR #187, opened 2016-10-28
Case-insensitive header matching

### LINKED ISSUE #184: Handle header matching in a case-insensitive way
There are a few places in the code where we check the headers against some value (mostly to check if the request was a JSON API one).  Whenever we do this, we should do so in a case-insensitive way.

XaserAcheron:
Closes #184 -- tests included, all passing on my end. Covers everything on [the checklist](https://github.com/ember-cli/ember-ajax/pull/119#issuecomment-226352391), plus a bit more.

I opted for a regex check for the JSON API content type solely to get rid of `indexOf`, because I am a silly person.

### DIFF
--- a/addon/mixins/ajax-request.js
+++ b/addon/mixins/ajax-request.js
@@ -618,7 +618,7 @@ export default Mixin.create({
     let shortenedPayload;
     const payloadContentType = getHeader(headers, 'Content-Type') || 'Empty Content-Type';
 
-    if (payloadContentType === 'text/html' && payload.length > 250) {
+    if (payloadContentType.toLowerCase() === 'text/html' && payload.length > 250) {
       shortenedPayload = '[Omitted Lengthy HTML]';
     } else {
       shortenedPayload = JSON.stringify(payload);
--- a/addon/mixins/ajax-request.js
+++ b/addon/mixins/ajax-request.js
@@ -43,13 +43,13 @@ const {
   testing,
   warn
 } = Ember;
-const JSONAPIContentType = 'application/vnd.api+json';
+const JSONAPIContentType = /^application\/vnd\.api\+json/i;
 
 function isJSONAPIContentType(header) {
   if (isNone(header)) {
     return false;
   }
-  return header.indexOf(JSONAPIContentType) === 0;
+  return !!header.match(JSONAPIContentType);
 }
 
 function startsWithSlash(string) {
--- a/addon/mixins/ajax-request.js
+++ b/addon/mixins/ajax-request.js
@@ -22,6 +22,7 @@ import {
   isSuccess
 } from '../errors';
 import parseResponseHeaders from '../utils/parse-response-headers';
+import getHeader from '../utils/get-header';
 import { RequestURL } from '../utils/url-helpers';
 import ajax from '../utils/ajax';
 
@@ -236,7 +237,7 @@ export default Mixin.create({
       url: hash.url
     };
 
-    if (isJSONAPIContentType(hash.headers['Content-Type']) && requestData.type !== 'GET') {
+    if (isJSONAPIContentType(getHeader(hash.headers, 'Content-Type')) && requestData.type !== 'GET') {
       if (typeof hash.data === 'object') {
         hash.data = JSON.stringify(hash.data);
       }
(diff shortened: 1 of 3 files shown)


addon/mixins/ajax-request.js
```diff
@@ -615,9 +616,9 @@ export default Mixin.create({
    */
   generateDetailedMessage(status, headers, payload, requestData) {
     let shortenedPayload;
-    const payloadContentType = headers['Content-Type'] || 'Empty Content-Type';
+    const payloadContentType = getHeader(headers, 'Content-Type') || 'Empty Content-Type';
 
-    if (payloadContentType === 'text/html' && payload.length > 250) {
+    if (payloadContentType.toLowerCase() === 'text/html' && payload.length > 250) {
```
XaserAcheron: Case-insensitive content-type check. This is technically "extra" but related-enough that I couldn't help but address it as well.

tests/unit/utils/get-header-test.js
```diff
@@ -0,0 +1,44 @@
...
+
+describe('getHeader', function() {
+  it('returns undefined when headers are undefined', function() {
+    let header = getHeader(undefined);
+
+    equal(header, undefined, 'undefined is returned');
+  });
+
+  it('returns undefined when name is undefined', function() {
+    let header = getHeader({}, undefined);
```
alexlafroscia: Another example of `let`; I won't mark them all.

tests/unit/utils/get-header-test.js
```diff
@@ -0,0 +1,44 @@
+/* jshint expr:true */
+import getHeader from 'ember-ajax/utils/get-header';
+import { describe, it } from 'mocha';
+import { assert } from 'chai';
```
alexlafroscia: Would you mind using `expect` instead? The codebase might not be there right now, but I'd like move in that direction.

tests/unit/utils/get-header-test.js
```diff
@@ -0,0 +1,44 @@
+/* jshint expr:true */
```
alexlafroscia: Drop these comments, we're using `eslint`

addon/mixins/ajax-request.js
```diff
@@ -615,7 +616,7 @@ export default Mixin.create({
    */
   generateDetailedMessage(status, headers, payload, requestData) {
     let shortenedPayload;
-    const payloadContentType = headers['Content-Type'] || 'Empty Content-Type';
+    const payloadContentType = getHeader(headers, 'Content-Type') || 'Empty Content-Type';
```
alexlafroscia: Is there some default `content-type` value that is assumed and should be used? Or is `Empty Content-Type` appropriate?

jherdman: IIRC text/plain

tests/unit/utils/get-header-test.js
XaserAcheron: Leftover from the `ember generate` process. Wish it were a simple thing to `extend` blueprints, but alas, [the issue is open](https://github.com/ember-cli/blueprint/issues/2).

XaserAcheron:
Thanks for the review! Doing the suggested edits at present.

addon/mixins/ajax-request.js
XaserAcheron: There isn't really a default. The standard in case the type is unknown to the sender is to [omit the Content-Type header entirely](http://stackoverflow.com/questions/1176022/unknown-file-type-mime/28652339#28652339), so the absence of a header is a valid "value" in its own right.

The string 'Empty Content-Type' just gets placed in a friendly error description field anyway, so I think it's fine. Unless we want to change the message to a more explicit 'No Content-Type Header' or somesuch; I'd be down with that.

jherdman: TIL! Many thanks for sharing that.

addon/utils/get-header.js
```diff
@@ -0,0 +1,24 @@
...
+ * @param {Object} headers
+ * @param {string} name
+ * @return {string}
+ */
+export default function getHeader(headers, name) {
+  if (isNone(headers) || isNone(name)) {
+    return; // ask for nothing, get nothing.
+  }
+
+  let matchedKey = A(Object.keys(headers)).find((key) => {
```
XaserAcheron: Fixed my additions, but after switching on `prefer-const`, ESLint is raising flags elsewhere in the code.  Mind if I do the config change + compliance fixes in a separate PR?

addon/mixins/ajax-request.js
alexlafroscia: Sounds good to me, thanks for the additional info @XaserAcheron

XaserAcheron: Glad to help! This stuff's all pretty mysterious at first; I'm always learning too.

XaserAcheron:
Made the suggested fixes, but I'm holding off on adding `prefer-const` to `.eslintrc`. I'd like to do a separate PR on that one since enabling it raises flags elsewhere in the code.

alexlafroscia:
Great, looks 👍 to me! Maybe I can get one more review before pulling this in?

@jherdman @taras

Issue #193, opened 2016-11-04
Setting a different host with no namespace is ignored


pijey:
Hi !

I think there is a problem when injecting the ajax service in a controller for instance and using directily when setting up a different host without namespace.
Exemple : 
- host : "http://localhost:8080"
- namespace : ""

When I call `this.get("ajax").request("my-request")...` I expect the request to hit the host I configured ("http://localhost:8080"). But the request hits the url **http://localhost:4200**.

In [this service](https://github.com/ember-cli/ember-ajax/blob/master/addon/mixins/ajax-request.js#L467) the regex always matches (because namespace is empty) and thus the request is made on the url **http://localhost:4200**.

As a workaround, I had to change my rest api to add a namespace i.e. "http://localhost:8080/api" and set up the ajax service this way : 
- host : "http://localhost:8080"
- namespace : "api"

jherdman:
Hi pijey. Would you be able to provide a reproduction app, or a failing test?

alexlafroscia:
Man, this host/namespace code is so problematic.  One of the big reasons I want to refactor this and come up with a better approach for Ember AJAX 3.

I'll add a failing test case with this configuration and work from there to fix this.

pijey:
Ok great ! Sorry, I don't have much time these days to provide a failing test

alexlafroscia:
I can't reproduce (see linked commit above). I'm going to close this for now -- if you can re-open with a reproduction, I'm happy to fix it!

PR #197, opened 2016-11-11
Test refactoring

alexlafroscia:
Misc stuff. Started as a response to #193, writing a test for it, but turned into a bunch of little changes.


bower.json
```diff
@@ -1,14 +1,22 @@
...
-    "mocha": "~2.2.4",
-    "chai": "~2.3.0",
-    "ember-mocha-adapter": "~0.3.1"
+    "mocha": "~2.2.4"
   },
   "resolutions": {
     "ember": "~2.5.0"
-  }
+  },
+  "version": "2.5.2",
```
jherdman: IIRC you don't want this stuff here as we're not publishing this as a bower installable dependency

XaserAcheron: I'd shy away from it personally because, knowing myself, I'd remember to update it in one file but forget to do the other. Depends on how goldfish-like everyone's memory is.

alexlafroscia: Fair point; I'll remove htat.

PR #199, opened 2016-11-23
Refactor url building

### LINKED ISSUE #180: Clean up URL building code
There's a few problems that I have with the way we handle building up the URL that I would like to address.
- I wrote the [`RequestURL` helper class](https://github.com/ember-cli/ember-ajax/blob/3ebfef02d37614c7f09d0d7d92ba741a47866664/addon/utils/url-helpers.js#L78) and think that's is likely over-engineered.  I'd rather refactor this toward a set of pure function helpers than have this as a class.
- We attempt to avoid adding the host name and namespace if they already exist in the base URL that gets passed.  This is done in order to support Ember Data, which passes us a URL that has those parts.  It ends up in some [really gross code that is pretty hard to ensure is correct](https://github.com/ember-cli/ember-ajax/blob/3ebfef02d37614c7f09d0d7d92ba741a47866664/addon/mixins/ajax-request.js#L472)
### Removing the `RequestURL` class
- [x] [Add a function for checking if a URL is a "complete" URL (includes the `http(s)://` already)](https://github.com/ember-cli/ember-ajax/blob/3ebfef02d37614c7f09d0d7d92ba741a47866664/addon/mixins/ajax-request.js#L454)
- [x] [Add a function for checking if two URLs have the same host](https://github.com/ember-cli/ember-ajax/blob/3ebfef02d37614c7f09d0d7d92ba741a47866664/addon/mixins/ajax-request.js#L601)
### General cleanup
- [ ] Avoid the need to check for the host and namespace by assuming that they're not included
  - This may take a change to Ember Data, or re-thinking how we integrate with it

alexlafroscia:
- Moves some of the utility code into the `-private` namespace
- Removes the `RequestURL` class
- Adds functions that provide the functionality of `RequestURL`

Closes #180


alexlafroscia:
@taras can you look this over? The coveralls "failure" isn't significant IMO and one of the two AppVeyor runs passed... I'm not sure why AppVeyor is so unstable ¯\_(ツ)_/¯

taras:
@alexlafroscia it looks great. 👍

PR #208, opened 2016-12-01
chore(package): dependency updates

alexlafroscia:


tests/helpers/start-app.js
```diff
@@ -2,13 +2,13 @@ import Ember from 'ember';
 import Application from '../../app';
 import config from '../../config/environment';
 
-const { merge, run } = Ember;
+const { assign, merge, run } = Ember;
```
jherdman: merge appears to be unused and is also deprecated.

package.json
```diff
@@ -25,17 +28,20 @@
...
   "license": "MIT",
+  "dependencies": {
+    "ember-cli-babel": "^5.1.5"
+  },
   "devDependencies": {
     "broccoli-asset-rev": "^2.4.3",
-    "coveralls": "^2.11.13",
-    "ember-cli": "2.9.1",
+    "coveralls": "^2.11.15",
+    "ember-cli": "^2.10.0",
```
jherdman: Huh, weird. I didn't think ember-cli was a floating version after running the upgrade procedure. Is this safe?

package.json
```diff
@@ -2,6 +2,9 @@
   "name": "ember-ajax",
   "version": "2.5.2",
   "description": "Service for making AJAX requests in Ember 1.13+ applications.",
+  "keywords": [
+    "ember-addon"
```
jherdman: Should we consider adding "ajax" to the keyword list whilst we're at it?

package.json
alexlafroscia: Hmm, interesting. I do the dependency updates for `ember-ajax` through Yarn, and that is what it changed it to by default.

jherdman: Very interesting. It makes me a little uncomfortable allowing flexibility on the minor bit given that ember-cli has an involved upgrade process. Perhaps we should consider nailing this down to an explicit version as per the usual npm-based upgrade process?

alexlafroscia: Sure, I'll re-pin it to the exact version. 

This really highlights how the CLI doesn't quite fit with Semver. A minor version *should not require changes* but always does.

PR #223, opened 2017-01-14
Add XHR property to promises

### LINKED ISSUE #67: Expose XHR
I was trying to follow the example pattern of aborting XHRs when retrying a task in ember-concurrency (http://ember-concurrency.com/#/docs/examples/autocomplete), but realized that there's no way to abort a request made by ember-ajax. Is exposing the underlying raw XHR per request a possibility?

alexlafroscia:
A long-standing (_really_ long-standing) request of `ember-ajax` has been to provide a better way to access the XHR object created by a request without needing to use `.raw`, which only gives you the object after the fact.

This PR adds a `.xhr` object to the promise returned by making a request that can be used to inspect or manipulate the request.  This is especially useful for aborting the request after it has been kicked off, which previously has been impossible.

Closes #67

### DIFF
--- a/addon/mixins/ajax-request.js
+++ b/addon/mixins/ajax-request.js
@@ -25,6 +25,7 @@ import parseResponseHeaders from 'ember-ajax/-private/utils/parse-response-heade
 import getHeader from 'ember-ajax/-private/utils/get-header';
 import { isFullURL, parseURL, haveSameHost } from 'ember-ajax/-private/utils/url-helpers';
 import ajax from 'ember-ajax/utils/ajax';
+import AJAXPromise from 'ember-ajax/-private/promise';
 
 const {
   $,
@@ -32,7 +33,6 @@ const {
   Error: EmberError,
   Logger,
   Mixin,
-  RSVP: { Promise },
   Test,
   deprecate,
   get,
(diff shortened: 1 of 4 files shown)


addon/mixins/ajax-request.js
```diff
@@ -225,15 +225,20 @@ export default Mixin.create({
...
+    const internalPromise = this._makeRequest(hash);
+
+    const ajaxPromise = new AJAXPromise((resolve, reject) => {
+      internalPromise.then(({ response }) => {
+        resolve(response);
+      })
+      .catch(({ response }) => {
+        reject(response);
+      });
     }, `ember-ajax: ${hash.type} ${hash.url} response`);
```
alexlafroscia: @taras How important do you think this extra Promise label is?  This method could be simplified to just:

```js
return this.raw(...arguments)
  .then(({ response }) => response)
  .catch(({ response }) => {
    throw response;
  });
```

without it.  The internal request promise will still have the label; is adding `response` to the end of it really that useful?

alexlafroscia: @taras How important do you think this extra Promise label is?  This method could be simplified to just:

```js
return this.raw(...arguments)
  .then(({ response }) => response)
  .catch(({ response }) => {
    throw response;
  });
```

without it.  The internal request promise will still have the label; is adding `response` to the end of it really that useful?

alexlafroscia: Oh yeah, I don't doubt that the debugging help here is great.  However, we already add a label to the promise [here](https://github.com/alexlafroscia/ember-ajax/blob/a7094e5c32a68a8678c4e19d2abf807aea9c1711/addon/mixins/ajax-request.js#L335).  So, we create a second promise here just to add an additional label to the promise, when we could just as easily leave the additional promise and the label out.

alexlafroscia: I think we should leave this be for this PR and consider simplifying things here as a separate matter.

taras: It looks fine the way it is.

alexlafroscia:
@taras if you're good with the changes here, can you approve the PR?

sukima:
@taras bumping you again. Lots of us are anxious for this.

taras:
I thought I did this a week ago. Sorry for the delay. It looks great

alexlafroscia:
You did @taras but I was looking/hoping for the "official" PR approval (like @jherdman did, top-right corner of the PR). 

Since you gave the verbal 👌 I'll merge this.

PR #226, opened 2017-01-16
Clone options object before modifying it

gnapse:
The `options` object could be [frozen](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Object/freeze) and therefore not modifiable. It is safer to work with a clone of it. This PR does just that.

If for some reason it is not acceptable to assign to the `options` argument, let me know and I can modify this accordingly.

### DIFF
--- a/addon/mixins/ajax-request.js
+++ b/addon/mixins/ajax-request.js
@@ -452,7 +452,7 @@ export default Mixin.create({
    * @return {Object}
    */
   options(url, options = {}) {
-    options = Ember.merge({}, options);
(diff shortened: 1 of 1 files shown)


alexlafroscia:
Great points, thanks @gnapse!

alexlafroscia:
Hmm, I know I already merged it but would you mind adding a test case for this in a separate PR @gnapse?

gnapse:
Sure @alexlafroscia, I just did. Check out #227.

PR #231, opened 2017-01-23
Ensure that `requestData` has the correct method and type

alexlafroscia:


alexlafroscia:
It's pretty silly that we use this `requestData` object at all, IMO. We only use it to create the promise label, which we could just do from original `hash` argument.  `handleResponse` could be called with the original XHR, which would eliminate the need for passing in the status text directly as well. Definitely a breaking change -- worth cleaning up for `3.0.0`?

cc: @jherdman @taras

alexlafroscia:
Hmm, so now there's a question of how to get this out into a release. `master` already has breaking changes.... Maybe we can branch from the last release, cherry-pick this and release again?

PR #232, opened 2017-01-25
Fixes #158. Don't modify null, undefined payload

### LINKED ISSUE #158: `null` response from server is returned as empty object
Upgrading to 2.5.0 from 0.7.1. #126 breaks my app, I expect a `null` value from the API, but now this returns an empty object, which breaks all my `{{#if data}}`. cc @alexlafroscia @taras

manoharank:
As per @taras [comment](https://github.com/ember-cli/ember-ajax/issues/158#issuecomment-246175431)

### DIFF
--- a/addon/mixins/ajax-request.js
+++ b/addon/mixins/ajax-request.js
@@ -535,8 +535,6 @@ export default Mixin.create({
    * @return {Object | AjaxError} response
    */
   handleResponse(status, headers, payload, requestData) {
-    payload = (payload === null || payload === undefined) ? {} : payload;
-
     let error;
 
     if (this.isSuccess(status, headers, payload)) {
@@ -849,6 +847,8 @@ export default Mixin.create({
    * @return {Array} An array of JSON API-formatted error objects
    */
   normalizeErrorResponse(status, headers, payload) {
+    payload = isNone(payload) ? {} : payload;
+
     if (isArray(payload.errors)) {
       return payload.errors.map(function(error) {
         if (isObject(error)) {
--- a/tests/unit/mixins/ajax-request-test.js
+++ b/tests/unit/mixins/ajax-request-test.js
@@ -837,7 +837,7 @@ describe('Unit | Mixin | ajax request', function() {
     });
   });
 
-  it('it doesn\'t reassign payloads which evaluate falsey but are not null or undefined', function() {
+  it("it doesn't reassign payloads which evaluate falsey", function() {
     const service = new AjaxRequest();
 
     const payloadWithFalseyString = service.handleResponse(200, {}, '');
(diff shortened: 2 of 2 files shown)


tests/unit/mixins/ajax-request-test.js
```diff
@@ -850,10 +850,10 @@ describe('Unit | Mixin | ajax request', function() {
     expect(isNaN(payloadWithNaN)).to.be.ok;
 
     const payloadWithNull = service.handleResponse(200, {}, null);
-    expect(payloadWithNull).to.be.deep.equal({});
+    expect(payloadWithNull).to.be.null;
 
     const payloadWithUndefined = service.handleResponse(200, {}, undefined);
-    expect(payloadWithUndefined).to.be.deep.equal({});
+    expect(payloadWithUndefined).to.be.unefined;
```
alexlafroscia: `unefined` -> `undefined`

tests/unit/mixins/ajax-request-test.js
```diff
@@ -837,7 +837,7 @@ describe('Unit | Mixin | ajax request', function() {
     });
   });
 
-  it('it doesn\'t reassign payloads which evaluate falsey but are not null or undefined', function() {
+  it('it doesn\'t reassign payloads which evaluate falsey', function() {
```
alexlafroscia: Can you double-quote the outer string instead of escaping the `'` here?

addon/mixins/ajax-request.js
```diff
@@ -849,6 +847,8 @@ export default Mixin.create({
    * @return {Array} An array of JSON API-formatted error objects
    */
   normalizeErrorResponse(status, headers, payload) {
+    payload = isNone(payload) ? {} : payload;
```
alexlafroscia: So, in `3.0.0` we're completely removing the implementation of `normalizeErrorResponse` in favor of just leaving it as a function that returns `payload`.  Given that, should we really be making changes to this method now?

manoharank: Otherwise, an error will happen in the next line `isArray(payload.errors)`.

manoharank: I have fixed other two reviews

manoharank:
Ping @alexlafroscia

manoharank:
@alexlafroscia any chance to merge this PR?

alexlafroscia:
I'm not sure how I feel about this making it into a 2.X branch. Isn't this a breaking API change? 

I'd love your thoughts @XaserAcheron  and/or @taras

XaserAcheron:
My two cents: I'd categorize it as a breaking change as it looks like the previous behavior was intentional, judging from the tests -- i.e. this isn't really a "bugfix" per se.

master is v3.0.0, though, so this PR ought to be good if all the requested changes are in.

alexlafroscia:
Considering the fact that `master` is v3, and v3 is basically going to no-op `normalizeErrorResponse`... Shouldn't we not do this, and instead just land the code change that removed the method in it's entirety? I thought that's what we had decided on at one of those weekly meetings...

XaserAcheron:
The PR is also modifying `handleResponse`, which is the more important bit.

The `normalizeErrorResponse` change is just there to prevent `if isArray(payload.errors)` from blowing up if payload is null/undefined (which would be allowed post-change).

No harm in adding the change IMO, then taking out `normalizeErrorResponse` entirely in a later commit.

alexlafroscia:
Gotcha. You've convinced me. Sorry I sat on this so long @manoharank

PR #239, opened 2017-02-17
Better URL detection for .get error

alexander-alvarez:
I tried doing `.get('some/url')` and was scratching my head for a while..

And that I know of people don't do `object.get('my/property/with/slashes')`


alexlafroscia:
Cool, thanks for the improvements! I'll restart the build that failed, it doesn't look like a problem with your code.

PR #240, opened 2017-02-17
Stringify all json by any content-type

mwpastore:
* Stringify application/json in additon to application/vnd.api+json
* Inspect the top-level contentType in addition to the Content-Type header

### DIFF
--- a/addon/mixins/ajax-request.js
+++ b/addon/mixins/ajax-request.js
@@ -45,7 +45,7 @@ const {
   testing,
   warn
 } = Ember;
-const JSONAPIContentType = /^application\/vnd\.api\+json/i;
+const JSONContentType = /^application\/(?:vnd\.api\+)?json/i;
 
 function defineDeprecatedErrorsProperty(error, errors) {
   Object.defineProperty(error, 'errors', {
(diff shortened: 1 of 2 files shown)


jherdman:
I'm definitely amenable to this. Would you mind adding some tests though?

addon/mixins/ajax-request.js
```diff
@@ -268,7 +268,13 @@ export default Mixin.create({
     const method = hash.method || hash.type || 'GET';
     const requestData = { method, type: method, url: hash.url };
 
-    if (isJSONAPIContentType(getHeader(hash.headers, 'Content-Type')) && requestData.type !== 'GET') {
+    if (
+      requestData.type !== 'GET'
```
alexlafroscia: I don't love this style of breaking the conditional into multiple lines... If you don't want to leave it inline, could you mind making it a function call instead?

mwpastore: I should probably break it out into its own npm package. 😉 Sure, I'll inline it.

mwpastore:
@jherdman Will do!

mwpastore:
Okay @alexlafroscia and @jherdman, please take another look and let me know if you'd like any further changes. Thank you!

alexlafroscia:
Ugh, I should really just disable Coveralls. Failing this because of a `0.2%` decrease is ridiculous.

alexlafroscia:
Thanks @mwpastore!

mwpastore:
You're welcome! Thank you!

PR #250, opened 2017-03-03
make sure contentType is a string before running a regex match

robclancy:
This caused issues when we are doing `POST` requests for uploads which we may or may not specify a `contentType`. If `contentType` is false it still tries to run `.match` and you get an error.

### DIFF
--- a/addon/mixins/ajax-request.js
+++ b/addon/mixins/ajax-request.js
@@ -71,7 +71,7 @@ function defineDeprecatedErrorsProperty(error, errors) {
 }
 
 function isJSONContentType(header) {
-  if (isNone(header)) {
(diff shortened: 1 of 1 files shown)


alexlafroscia:
Seems reasonable. Can you add a test?

Also, for my own knowledge, what's the use-case on `contentType` being `false`?

robclancy:
I already explained the use case. If I don't want to set a content type header but sometimes do, using false is the logical thing when you don't want to use it (or well, falsey).

I tried adding a test at first but it really seemed pointless to add a test for types passed to such method.

If I add tests it won't be for some time as I have no time leading up to a vacation (preparing others).

PR #256, opened 2017-03-20
Response processing example formatted incorrectly

ctataryn:
I believe the use of `{ ... }` were meant to be markdown formatting, however they rendered in the code example instead. I removed them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -288,8 +288,8 @@ If you need to access the json response of a request that failed, you can use th
 
 ```js
 this.get('ajax').raw(url, options)
-  .then(({ response }) => this.handleSuccess(response))
-  .catch(({ response, jqXHR, payload }) => this.handleError(response));
+  .then((response) => this.handleSuccess(response))
+  .catch((response, jqXHR, payload) => this.handleError(response));
 ```
 
 Note that in this use case there's no access to the error object. You can inspect the `jqXHR` object for additional information about the failed request. In particular `jqXHR.status` returns the relevant HTTP error code.


webark:
@ctataryn that looks more like they where wanting to suggest using restructuring. right? I'd have to look at the object that get passed to those to be sure though.

alexlafroscia:
@webark has it right; in those code examples, the `{ response }` denotes that we're destructuring that property off of the return value.

Promises can only resolve a single value, but in our case, we have multiple things that we want to provide as part of the "return" value.  So, we can bundle them up as a POJO on one end, resolve that object, and then de-structure those values out the other side.

You can see an example of creating that object [here](https://github.com/ember-cli/ember-ajax/blob/09fa3f060c244021c52157c34ea1894c00ba908c/addon/mixins/ajax-request.js#L341), and an internal example of destructuring the properties [here](https://github.com/ember-cli/ember-ajax/blob/09fa3f060c244021c52157c34ea1894c00ba908c/addon/mixins/ajax-request.js#L246).

Thanks for the PR though, @ctataryn; I appreciate the vigilance in trying to keep our docs clear!

ctataryn:
@alexlafroscia @webark thanks for the clarification. I did what any programmer worth their weight does: I blindly copy/pasted the example from the README into my own code without checking the context :trollface: 

When it didn't work (i.e. `response` was undefined because this was a success, not a failure) I removed the braces and voila.  Thanks to your links I now understand the syntax of destructuring the argument, thanks for that.

alexlafroscia:
No problem. Your issue _might_ have been `.request` vs `.raw`.

`.request` basically just calls `.raw`, automatically destructures the `response` property and then resolves _that_, since it's like 99% of the time what you need.  For that other 1% of the time, you can call `.raw` which does the same logic, but provides a lot more information (response, XHR, status code, etc).

PR #265, opened 2017-03-27
Simplify `normalizeErrorResponse`

### LINKED ISSUE #264: Make normalizeErrorResponse only return payload without modifying the response

alexlafroscia:
Over time, we decided that the existing `normalizeErrorResponse`
behavior tried to do way too much.  We've decided that the default
behavior should be to instead just return the error payload that the
server provides, without making modifications to it.

The old behavior can be re-introduced by including the mixin found at
`mixins/legacy/normalize-error-response`.

This also removes the deprecation warnings around the `errors` property
in favor of the replacement `payload` property.

Closes #264

### DIFF
--- a/addon/mixins/ajax-request.js
+++ b/addon/mixins/ajax-request.js
@@ -46,29 +46,6 @@ const {
 } = Ember;
 const JSONContentType = /^application\/(?:vnd\.api\+)?json/i;
 
-function defineDeprecatedErrorsProperty(error, errors) {
-  Object.defineProperty(error, 'errors', {
-    get() {
-      deprecate(
-        'This property will be removed in ember-ajax 3.0.0. Please use `payload` going forward. Note the attached URL for details.',
-        false,
-        {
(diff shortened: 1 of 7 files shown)


taras:
@alexlafroscia this looks good. This should target 3.0.0. Right?

alexlafroscia:
Yes; I'm going to merge into master and cut a `3.0.0-beta.1` from master once it lands

alexlafroscia:
Actually, once this _and_ #266 land (just need a passing build on Travis for that one) then I'll cut a 3.0.0

alexlafroscia:
Okay, well I'm going to go ahead and assume the verbal " 👍 " from @taras means it's safe to pull this in.

XaserAcheron:
Aces! Glad to see some serious momentum on this as of late. O:

Sorry I haven't been doing the same -- things got real busy  on both work- and non-work fronts.

alexlafroscia:
I have too. EmberConf inspired to me stop dragging my feet though 😆

PR #266, opened 2017-03-27
Update to use Babel 6

alexlafroscia:


alexlafroscia:
Issue created about the failing build here:

https://github.com/babel/ember-cli-babel/issues/129

Maybe this can make it into `3.0.0-beta.1`

stefanpenner:
It appears the failure was in this project. I have updated this pr:

* use local `require`
* correctly detect node <-- this was the fix
* update lock file <-- just case :P

PR #275, opened 2017-04-13
Expose status codes on error objects

bendemboski:
There may be cases where applications want access to the status code that caused the error (e.g. logging/metrics, or terse user-facing error messages), so this allows them to easily get at it.

My use case is two-fold:

1. I use mixpanel to log metrics about failed requests that include the status code as a property
1. When I encounter an unhandled error, I don't want to try to communicate the distinction between "bad request", "server error", etc. -- I just want to show them a generic "Oops, my bad" message with the status code in parentheses so we have something to go on if they contact support.

Given how difficult it is to recover the status code from an error object (especially if it's just an `AjaxError` -- then the only way is to run a regex on the message ☹️ ), this seems like something worth including in the base implementation and not requiring users to subclass the service in order to get.

### DIFF
--- a/addon/mixins/ajax-request.js
+++ b/addon/mixins/ajax-request.js
@@ -555,6 +555,7 @@ export default Mixin.create({
 
       error = new AjaxError(payload, detailedMessage);
     }
+    error.status = status;
 
     return error;
   },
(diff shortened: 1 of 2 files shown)


bendemboski:
Yeah, that all definitely makes sense to me. I think I'll have some time this weekend to attack this (or if you have it all in your head and want to take it over yourself, I won't be put off or anything).

alexlafroscia:
Sweet. Give it a shot if you're able, I've got some other OSS stuff on my plate for the weekend.

Kilowhisky:
+1 the new error response types hides status codes.  Doing type checking is excessive for unnecessary reasons.

alexlafroscia:
#318 was just merged that handles this change, so I'm going to go and close this. Thanks for the PR though!

Issue #177, opened 2016-10-19
Drop AJAX error detection helpers


alexlafroscia:
I feel like the error detection functions that we provide aren't all that useful.  Specifically:
- Functions should generally take one kind of input and return one kind of output.  We take either an `AjaxError` instance _or_ a number and try to verify that it's describing some kind of error. While useful, I think this is kind of confusing behavior.
- People really can't (or shouldn't) have code where the argument is some variable that could be either an error object or a number; they must know ahead of time whether it's a number or error object and could just check it directly
- If they are using it with only one argument type in mind, then they should just do a number comparison, or an `instanceof` check, themselves.  Preferably the latter, since our descriptive error classes remove the need checking the status code directly.

If we were to remove them in `v3.0.0`, then for the time being we would deprecate the error checkers, and then remove them entirely in `v3.0.0`.

jherdman:
👍

sandstrom:
Somewhat related, it would be neat if it was easy to either `reopen` ajax errors, or provide a custom error base class.

alexlafroscia:
@sandstrom what's the use case on that?

sandstrom:
Our server responses contain a bunch of useful metadata, in a specific JSON format.

Instead of having to dig out details, e.g. `myError.payload.envelope.meta.request_id` I'd like to add a property to the error class such that `myError.requestId`would exist.

An alternative to providing a custom error class would be the ability to modify/setup the error instance after it's created. I know there is `parseErrorResponse` and `normalizeErrorResponse` but none of them allow setting up properties on the error instance itself.

alexlafroscia:
Huh, interesting. Definitely a reason to do this.

Right now, we use a regular JS class for the errors. If we adopted Ember classes, you'd be able to reopen them; would that fit your use case?

Also, I'd love to take this discussion to a new issue. Would you make one and explain the use case (coping you comment here is fine) @sandstrom?

sandstrom:
@alexlafroscia Done! https://github.com/ember-cli/ember-ajax/issues/292

alexlafroscia:
I've changed my mind on this and would rather just keep them around.

PR #293, opened 2017-05-02
Group HTTP errors

sandstrom:
Easier to detect if the error occurred on the HTTP level or elsewhere (abort, timeout).

@alexlafroscia What do you think about the general idea here? (I know tests are missing, etc)

I've changed some messages, but that may break if people do string matching on error messages (instead of using the helpers or instanceof testing). Should we consider that a breaking change? I can revert the message changes too.

***

I think it would make sense to hold off merging this, since there are [planned changes to the helpers](https://github.com/ember-cli/ember-ajax/issues/177) in 3.0.0.

### DIFF
--- a/addon/errors.js
+++ b/addon/errors.js
@@ -20,62 +20,72 @@ export function AjaxError(errors, message = 'Ajax operation failed') {
 AjaxError.prototype = Object.create(EmberError.prototype);
 
 /**
+ * @class HttpError
+ * @public
+ */
+export function HttpError(errors) {
+  AjaxError.call(this, errors, 'Http request failed');
+}
+
+HttpError.prototype = Object.create(AjaxError.prototype);
+
+/**
  * @class InvalidError
  * @public
  */
 export function InvalidError(errors) {
-  AjaxError.call(this, errors, 'Request was rejected because it was invalid');
+  HttpError.call(this, errors, 'Request was rejected because it was invalid');
 }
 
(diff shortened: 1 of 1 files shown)


alexlafroscia:
I'm definitely 👍  for these changes. The potentially breaking nature could be problematic though... Can you recreate one of those breaking instances in a test so I can get a better idea of what that would look like?

Also, the 3.0 release is already out, so we're looking at breaking changes for 4.0 instead. However, I have no problem releasing 4.0 soon if it helps us clean up the API a bit.

sandstrom:
It would be something like this:

```js
myXhr.then((response) => {
  // …
}).catch((err) => {
  if (err.toString() === 'Ajax authorization failed') {
    // …
  } else {
    // …
  }
});

// OR

myXhr.then((response) => {
  // …
}).catch((err) => {
  if (err.message === 'Ajax authorization failed') {
    // …
  } else {
    // …
  }
});

```

I think this is unlikely, since there are helpers precisely to avoid doing this. My guess is that ~99% of people will either use `isUnauthorizedError(err)` or `instanceof UnauthorizedError`.

alexlafroscia:
Gotcha. I like this change then.

Should we use `HTTPError` or `HttpError`? Could go either way.

alexlafroscia:
This is not going to merge well after the TypeScript rewrite; I'm going to close this in favor of a fresh PR in the future.

PR #318, opened 2017-10-16
Expose error codes

spruce:
As written in #275 it would make sense to expose the actual http status codes outside of the default error classes.
Only thing I can see which blocked the PR was this: https://github.com/ember-cli/ember-ajax/pull/275#pullrequestreview-32876092
So I tried to do everything of it. Only thing I couldn't achieve is to make it a switch statement. (I don't think it's possible, cause case expects only a value and not a expression and function calls are expressions. Right??? I'm really unsure but would like to know if I'm right or not)

Happy to work some more to get it merged 👀

### DIFF
--- a/addon/mixins/ajax-request.js
+++ b/addon/mixins/ajax-request.js
@@ -528,12 +528,11 @@ export default Mixin.create({
     // Allow overriding of error payload
     payload = this.normalizeErrorResponse(status, headers, payload);
 
-    return _createCorrectError(status, headers, payload, requestData);
+    return this._createCorrectError(status, headers, payload, requestData);
   },
 
   _createCorrectError(status, headers, payload, requestData) {
     let error;
-
 
     if (this.isUnauthorizedError(status, headers, payload)) {
       error = new UnauthorizedError(payload);
--- a/addon/mixins/ajax-request.js
+++ b/addon/mixins/ajax-request.js
@@ -551,6 +551,7 @@ export default Mixin.create({
 
       error = new AjaxError(payload, detailedMessage);
     }
+    error.status = status;
 
     return error;
   },
--- a/addon/mixins/ajax-request.js
+++ b/addon/mixins/ajax-request.js
@@ -521,14 +521,19 @@ export default Mixin.create({
    * @return {Object | AjaxError} response
    */
   handleResponse(status, headers, payload, requestData) {
-    let error;
-
     if (this.isSuccess(status, headers, payload)) {
       return payload;
     }
 
     // Allow overriding of error payload
     payload = this.normalizeErrorResponse(status, headers, payload);
+
+    return _createCorrectError(status, headers, payload, requestData);
+  },
+
+  _createCorrectError(status, headers, payload, requestData) {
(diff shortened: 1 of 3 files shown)


alexlafroscia:
Hey @spruce, sorry this has sat so long. I'm getting back into OSS now and would love to help get this landed.

To start, there's a merge conflict now. Mind taking a look at that?

tests/unit/mixins/ajax-request-test.js
```diff
@@ -879,6 +881,7 @@ describe('Unit | Mixin | ajax request', function() {
         .catch(function(reason) {
           expect(isTimeoutError(reason)).to.be.ok;
           expect(reason.payload).to.be.null;
+          expect(reason.status).to.equal(-1);
```
alexlafroscia: Where did this `-1` come from?

addon/mixins/ajax-request.js
```diff
@@ -520,15 +520,19 @@ export default Mixin.create({
...
       return payload;
     }
 
     // Allow overriding of error payload
     payload = this.normalizeErrorResponse(status, headers, payload);
 
+    return this._createCorrectError(status, headers, payload, requestData);
+  },
+
+  _createCorrectError(status, headers, payload, requestData) {
```
alexlafroscia: What's the reason for breaking this out into a separate method, rather than just adding the logic to the existing `handleResponse`?

spruce:
Will work on it. Would still like for this to get merged. I'm going to work on this now.

tests/unit/mixins/ajax-request-test.js
spruce: It comes from here: https://github.com/ember-cli/ember-ajax/pull/318/files#diff-6b893c2f1942ff5e1a6a907d293c60e3R83

addon/mixins/ajax-request.js
spruce: I actually came from a request you had: https://github.com/ember-cli/ember-ajax/pull/275#pullrequestreview-32876092 
Or did I misunderstand something there?

alexlafroscia: That's fair 😅 That review was so long ago. I'm fine with merging as-is. I'm working on a pretty major refactor so might clean it up later if it's bugging me.

tests/unit/mixins/ajax-request-test.js
alexlafroscia: 👍

mehulkar:
@spruce @alexlafroscia Thanks for making getting this in! Could we get it into a patch release? :). I want my 502/503 codes :D

alexlafroscia:
Yeah, I _think_ I should be able to cut a release, it doesn't seem like there's anything that's landed between then and now that would be breaking. Sorry this has lingered so long, I didn't realize the last major release was back in April!

alexlafroscia:
https://github.com/ember-cli/ember-ajax/releases/tag/v3.1.0

mehulkar:
@alexlafroscia thank you!

PR #344, opened 2018-03-20
Fix URL generation for paths with namespace as prefix

Charizard:


alexlafroscia:
Thanks for this! I'm going to merge this even with the failing build because I have plans to fix that issue over the weekend. Once the build is passing again, I'll merge cut a release with these changes.

PR #356, opened 2018-06-18
Convert to TypeScript

### LINKED ISSUE #346: Interest in TypeScript?
Hello! I use your addon and really appreciate it. I'm also participating in [this quest issue to add TypeScript support throughout the Ember.js ecosystem](https://github.com/typed-ember/ember-typings/issues/14). (This will benefit to anyone who uses your addon, not just TypeScript users!) Are you interested in either converting the addon to use TypeScript itself, or in adding type definitions? I already have worked out most of the type definitions in my own project, which I could provide through [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped) instead. Thanks!

alexlafroscia:
- Re-writes the logic in TypeScript
- Drops many of the JSDoc comments in favor of the explicit TypeScript typings

Closes #346 

### Todo

- [x] Fix service registry entry


alexlafroscia:
It's TBD on whether this will need to be a breaking change or not. I'm converting the `AjaxError` classes over to _actual_ classes, which may not work on all Ember versions.

Since that'll make maintenance easier, I think I'm OK with supporting Ember 2.4+ or something like that, if it means we can use some of the modern features in Ember. I don't really feel like active 1.13 maintenance is really necessary any more.

addon/services/ajax.ts
```diff
@@ -0,0 +1,11 @@
+import Service from '@ember/service';
+import AjaxRequestMixin from '../mixins/ajax-request';
+
+export default Service.extend(AjaxRequestMixin);
+
+// DO NOT DELETE: this is how TypeScript knows how to look up your services.
+declare module '@ember/service' {
+  interface Registry {
+    ajax: any;
```
alexlafroscia: I couldn't get this right, so I just set it to `any` for now.

If I export the service like this instead:

```typescript
export default class AjaxService extends Service.extend(AjaxRequestMixin);
```

some of my tests started complaining that I wasn't using `new` with the class.

types/ember-fastboot/global.d.ts
```diff
@@ -0,0 +1,12 @@
+import url from 'url';
+
+interface ModuleRegistry {
```
alexlafroscia: This bit was kind of weird and hard to figure out, but it allowed me to get around not having types for the `FastBoot` global yet.

alexlafroscia:
@bouke @chriskrycho Would you guys look this over?

addon/mixins/ajax-request.ts
```diff
@@ -333,81 +359,48 @@ export default Mixin.create({
...
    * calls `request()` but forces `options.type` to `POST`
-   *
-   * @method post
-   * @public
-   * @param {string} url The url to make a request to
-   * @param {Object} options The options for the request
-   * @return {Promise} The result of the request
    */
-  post(url, options) {
+  post(url: string, options?: AJAXOptions): AJAXPromise<Response> {
```
Bouke: See my typings where I have generic variants for all the verbs, so for example;

```
        post(url: string, options?: JQuery.AjaxSettings): RSVP.Promise<any>;
        post<T>(url: string, options?: JQuery.AjaxSettings): RSVP.Promise<T>;
```

This allows for clean code on the consumer's side, as the return type is already defined.

Bouke:
I tried adding a file, which ran into an error:

```
file added routes\bouke.ts
events.js:183
      throw er; // Unhandled 'error' event
      ^

Error: Error watching file for changes: EPERM
    at _errnoException (util.js:992:11)
    at FSEvent.FSWatcher._handle.onchange (fs.js:1359:9)

error Command failed with exit code 1.
```

addon/mixins/ajax-request.ts
alexlafroscia: Makes sense to me, but could you give an example of what that would look like in use? I thought about going that direction with the type definitions, but decided against it because the return value is always going to be some kind of JSON object -- does this just allow the user to say something like this?

```typescript
interface User {
  name: string;
}

const user = await get<User>(this, 'ajax').get('/users/1');
```

and allow their application to know that `user` is a `string`, and give the right Intellisense?

My reservation here is that we have no way to know that `user` is _actually_ matching the shape of the `User` interface; we could try to tell TypeScript that that's the case, but we have no way to verify it. I think it would actually be safer to keep the `AJAXPromise<Response>` return type where `Response` is `any` because it's closer to what we can actually guarantee is true about the system.

alexlafroscia: On the other hand, we _could_ take a "user at your own risk" attitude toward the generic type with something like this:

```typescript
post<T = Response>(url: string, options?: whatever): AjaxPromise<T>
```

since that would basically tell people that, if left unspecified, the return value is going to resolve to `any`. However, if they _want_ to tell the compiler what the type should be treated as, they can -- at their own risk, because they may actually be wrong and cause a runtime error that TypeScript can't help them with.

I think I like this idea best, since it is kind of the "best of both worlds" -- if a user wants to opt into this feature, they can, or else fall back to the original behavior I described, which is arguably safer but might be less helpful.

alexlafroscia:
What do you mean, by trying to add a file? I'm not sure what you mean.

Bouke:
> What do you mean, by trying to add a file? I'm not sure what you mean.

Sorry, must've been the early hour -- I tried to reply to a different issue in a different repository.

alexlafroscia:
No problem!

addon/mixins/ajax-request.ts
alexlafroscia: Alright, that's what I did! Thanks for suggesting the generic.

I like the option of having it, if the user is confident in the structure of their JSON response and wants to get the richer experience, but by default it will be of type `Response` (which is an alias for `any`).

What's nice is that `request` and `raw` can both be parameterized by a type, even though `raw` actually has a return type of `AjaxPromise<RawRequest<T>>`. Adding in the generics was fun!

PR #365, opened 2018-07-18
fix: don't append leading '/' when building url

baljeet03:
Cases being handled here -
1) `{ host: "//", url: "/api/request" }` or `{ host: "//", url: "api/request" }`. It will generate `//api/request`
2) `{ url: "/api/request" }`. It will generate `/api/request`.
3) `{ url: "api/request" }`. It will generate `api/request`. No extra leading slash.
4) Same will happen in case of namespace. So `{ namespace: "/v1", url: "/api/request"}` will generate `/v1/api/request` and `{ namespace: "v1/", url: "/api/request" }` will generate `v1/api/request`.

### DIFF
--- a/addon/mixins/ajax-request.ts
+++ b/addon/mixins/ajax-request.ts
@@ -102,6 +102,10 @@ function removeLeadingSlash(string: string) {
   return string.substring(1);
 }
 
+function removeTrailingSlash(string: string) {
+  return string.slice(0, -1);
+}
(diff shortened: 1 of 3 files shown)


alexlafroscia:
This looks great, thanks for the PR! 

The tests are failing, but I've been having trouble with that build on `master` too so I'm going to go ahead and merge it in.

PR #380, opened 2018-09-11
Use `assign` instead of deprecated `merge`

wagenet:


addon/mixins/ajax-request.ts
```diff
@@ -434,16 +434,16 @@ export default Mixin.create({
    */
   _getFullHeadersHash(headers?: Headers): Headers {
     const classHeaders = get(this, 'headers');
-    const _headers = merge({}, classHeaders);
-    return merge(_headers, headers);
+    const _headers = assign({}, classHeaders);
+    return assign(_headers, headers);
```
rwjblue: Can you refactor this to a single assign call?

    const _headers = assign({}, classHeaders, headers);

wagenet: Done

alexlafroscia:
Released in `3.1.1`

PR #391, opened 2018-10-07
fix: skip instead of returning url when namespace is present in url

cosmosgenius:
When the namespace is present in the url, the host value being set is not taken into account. This PR fixes that issue. I have added a Test case for the scenario.
 
Let me know if anything else is required.


alexlafroscia:
This is great -- thanks so much for the PR!

PR #392, opened 2018-10-08
Fix is*Error type definitions

boris-petrov:
They should accept any type of `AjaxError` and not the specific type they check for as this doesn't make sense.


alexlafroscia:
Good catch! Thanks for the PR!

PR #432, opened 2019-02-28
Fix Node.js compatibility claim in package.json

boris-petrov:
`ember-ajax` requires and tests with Node.js 6+ but in `package.json` it says `4+`. This PR fixes that.


boris-petrov:
Added a compatibility section.

PR #433, opened 2019-02-28
Update ember-cli-babel to version 7

### LINKED ISSUE #431: ember-ajax does not work with Ember 3.8 when transpiling to ES5
[Original issue in Ember.js](https://github.com/emberjs/ember.js/issues/17643).

[Babel cannot transpile to ES5 extending from some built-in types](https://babeljs.io/docs/en/caveats/#classes) which means that [extending EmberError](https://github.com/ember-cli/ember-ajax/blob/3ed08ceb0b797a8022477edc8274e8db22c35c4b/addon/errors.ts#L3-L13) won't work any more in ES5 since Ember 3.8 because it is the [same as Error](https://github.com/emberjs/ember.js/pull/17216).

boris-petrov:
Fixes #431.

### DIFF
--- a/addon/errors.ts
+++ b/addon/errors.ts
@@ -1,6 +1,4 @@
-import EmberError from '@ember/error';
-
-export class AjaxError extends EmberError {
+export class AjaxError extends Error {
   payload: any;
   status: number;
 
--- a/README.md
+++ b/README.md
@@ -357,6 +357,10 @@ export default function someOtherUtility(url) {
 
 This information is only relevant if you're looking to contribute to `ember-ajax`.
 
+### Compatibility
+
+- Ember CLI v2.13 or above
+
 ### Installation
 
 - `git clone` this repository
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     ]
   },
   "dependencies": {
-    "ember-cli-babel": "^6.16.0",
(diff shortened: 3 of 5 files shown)


addon/errors.ts
```diff
@@ -1,6 +1,4 @@
-import EmberError from '@ember/error';
-
-export class AjaxError extends EmberError {
+export class AjaxError extends Error {
```
Turbo87: to be a bit more verbose here: Ember 3.8 changed `EmberError` to no longer be an `Error` subclass, but instead reexport `Error`. Babel 6 can't deal with extending native classes, but Babel 7 can. Babel 7 does not detect that it needs to do that though because of the reexport.

Since the bump to `ember-cli-babel@7` and the corresponding change in the minimum Ember CLI requirements require a major version bump anyway it seems fine to change the base class here to a regular `Error` too.

Turbo87:
merging now as it will fix the CI issues and unblock the other PRs

@alexlafroscia if you don't have time for it right now, I'm happy to handle the release, just need the npm permissions 😉

addon/errors.ts
runspired: you just solved why we couldn't figure out subclassing with that change properly in ember-data (not detecting the native class due to re-export)

Turbo87: happy to help 😉

PR #434, opened 2019-02-28
Drop old releases from Travis testing

boris-petrov:


boris-petrov:
I force-pushed with fixes for the comments.

Turbo87:
thinking some more about it, I'm not sure anymore if this is a good idea. there is apparently no immediate need for us to drop support for those Ember versions and it doesn't hurt us much to keep support for them other than a slightly longer CI time.

independent from dropping support for old Ember releases we should definitely add the newer ones to the test matrix though.

boris-petrov:
Well, this way we could make people update faster to newer versions. :) Otherwise yes, supporting older versions is fine. Should I make the change?

Turbo87:
@boris-petrov yeah, that would be good. I guess it's easiest to open a new PR for it.

the other PRs are now finally released as v5.0.0! 🎉

boris-petrov:
Closing in favor of https://github.com/ember-cli/ember-ajax/pull/438. Thanks a lot for the 5.0.0 release!

PR #440, opened 2019-03-22
fix: types on public properties in ajax-request mixin

jamescdavis:
Cast types of public properties initialized to `undefined` to `undefined | <type>` so that they can be overridden without type errors.


mike-north:
Merged because CI failure is clearly unrelated

PR #441, opened 2019-03-22
Tests for type information

mike-north:
We ran into some problems where type information was in misalignment with documentation (i.e., #440)

This PR sets up infrastructure for testing type information using [dtslint](https://github.com/Microsoft/dtslint).

Some fixes were needed in types for the library itself, but they all seem like improvements


addon/utils/ajax.ts
```diff
@@ -1,6 +1,6 @@
-import jQuery from 'jquery';
+import * as jQuery from 'jquery';
 
-const ajax =
+const ajax: typeof jQuery.ajax =
```
mike-north: because `FastBoot.require` is not well-typed, the type of `ajax` ended up being intersected with `any` (resulting in a type of `any` because `X | any` is `any` regardless of X)

type-tests/tests.ts
```diff
@@ -0,0 +1,44 @@
...
+  model() {
+    return this.get('ajax').request('/posts');
+  }
+});
+
+Ember.Controller.extend({
+  ajax: Ember.inject.service('ajax'),
+  actions: {
+    sendRequest() {
+      this.get('ajax').headers;
```
mike-north: This sets us up for a test that failed until #440 

```ts
// headers used to just be undefined
this.get('ajax').headers; // $ExpectType Headers | undefined
```

addon/utils/ajax.ts
```diff
@@ -1,6 +1,6 @@
-import jQuery from 'jquery';
+import * as jQuery from 'jquery';
```
mike-north: This does type information consumers a favor by allowing `ember-ajax` to work without forcing them to turn on `allowSyntheticDefaultImports` or `esModuleInterop` flags in their `tsconfig.json`. Libraries should generally _not_ enable these flags for this reason

mike-north: Looks like we may need to resolve this in a separate PR. What we have for now is tests that prove consumers may _need_ `"allowSyntheticDefaultImports": true` in their tsconfig in order to use anything that touches `addon/utils/ajax.ts` or the types it emits.

type-tests/tsconfig.json
```diff
@@ -0,0 +1,23 @@
+{
+  "compilerOptions": {
+    "module": "commonjs",
+    "lib": ["es6", "dom"],
+    "noImplicitAny": true,
+    "noImplicitThis": true,
+    "strictNullChecks": true,
+    "strictFunctionTypes": true,
+    "noEmit": true,
+    "allowSyntheticDefaultImports": true,
```
mike-north: This indicates that all consumers also need this flag. More info here: https://github.com/ember-cli/ember-ajax/pull/441#discussion_r268309448

PR #453, opened 2021-06-03
Migrate from Travis to GitHub Actions

boris-petrov:
By using:
https://github.com/jelhan/create-github-actions-setup-for-ember-addon
And adding a couple more linter steps from the Travis config file

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,197 @@
+name: CI
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+
+env:
+  NODE_VERSION: '6'
+
+jobs:
+  lint:
+    name: Lint
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        fetch-depth: 1
+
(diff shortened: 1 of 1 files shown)


boris-petrov:
@Turbo87 - here is an initial version. It doesn't seem to run for some reason - perhaps you'll have to enable it in [here](https://github.com/ember-cli/ember-ajax/actions) somehow.

Also, note that this probably will fail again. The CI uses Node 6 which is very old and perhaps that's one of the reasons Travis fails on the other PR.

rwjblue:
> It doesn't seem to run for some reason

GH Actions won't run from a fork

boris-petrov:
@rwjblue - so what do you suggest?

rwjblue:
@boris-petrov - Leave `.travis.yml` (they'll both run then) until we can confirm the GH Actions setup works

boris-petrov:
@rwjblue - did that but GitHub Actions doesn't seem to be running...

Turbo87:
@boris-petrov I tried it with your config, but unfortunately it failed in several way. I've merged a GitHub Actions config now in #459

PR #471, opened 2022-10-21
[DOCS] Add deprecation notice

bertdeblock:
/cc @chriskrycho

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,10 @@ Service for making AJAX requests in Ember applications.
 - improved error handling
 - ability to specify request headers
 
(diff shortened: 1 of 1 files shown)


webark:
could point people to https://github.com/expel-io/ember-ajax-fetch for an ease of transition for legacy apps. @rwwagner90

RobbieTheWagner:
@webark yeah, that sounds like a great idea 👍

bertdeblock:
Thanks for the suggestion. Updated accordingly.

kategengler (approved):

